PR #16, opened 2021-11-29
UI as type

Arpita-Jaiswal:


src/p2/interpreter.rs
```diff
@@ -12586,6 +12586,65 @@ mod test {
...
+            "foo/bar",
+            indoc::indoc!(
+                "
+                -- component foo: hello
+                component: ftd.text
+                color: red
+
+                -- component bar:
+                component: ftd.column
+                ftd.text $t:
```
amitu, line 12632: Lets only accept `ftd.ui`, not individual components.

src/p2/kind.rs
```diff
@@ -448,6 +449,10 @@ impl Kind {
                 "message" => Kind::Message,
                 "string-message" => Kind::StringMessage,
                 "int-message" => Kind::IntMessage,
+                "ftd.ui" => Kind::UI {
+                    name: "ftd.ui".to_string(),
```
amitu, line 453: Since it will always be `ftd.ui`, do we need this?

PR #24, opened 2021-12-07
`from_json_row` and `from_json_rows` added

Arpita-Jaiswal:


src/component.rs
```diff
@@ -2291,6 +2291,7 @@ mod test {
                 name: "foo/bar#person".to_string(),
                 fields: person_fields(),
                 instances: Default::default(),
+                order: vec![s("name"), s("address"), s("bio"), s("age")],
```
amitu, line 2294: You can maybe create `f("name, address,bio,age")` helper.

PR #33, opened 2021-12-27
Iteratively check the title heading as per priority

sharmashobhit:


amitu:
@sharmashobhit please publish on crates.io.

amitu:
After the test has passed. I just realised I did not wait for that.

PR #55, opened 2022-01-07
Bye Bye ftd.markup

Arpita-Jaiswal:


src/component.rs
```diff
@@ -1,3 +1,5 @@
+use crate::Element;
```
amitu, line 1: We don't do this at FifthTry :-p

PR #72, opened 2022-01-13
Import: exposing

Arpita-Jaiswal:
Now you can do this:

main.ftd
```ftd
-- import: record
exposing: *

-- import: lib
exposing: amitu, default-phone

-- ftd.boolean: $bool

-- ftd.text: $amitu.name

-- ftd.text: $default-phone

-- address:
```

record.ftd
```ftd
-- boolean bool: true

-- ftd.text address: Prayagraj
```

lib.ftd
```ftd
-- string default-phone: 1000

-- record person:
caption name:

-- person amitu: Amit Upadhyay
```

And you get output like:
```
true
Amit Upadhyay
1000
Prayagraj
```

### DIFF
--- a/src/p1/header.rs
+++ b/src/p1/header.rs
@@ -279,6 +279,21 @@ impl Header {
         }
     }
 
+    pub fn strings(&self, name: &str) -> Vec<&str> {
+        let mut strings = vec![];
+        for (_, k, v) in self.0.iter() {
+            if k.starts_with('/') {
+                continue;
+            }
+            if k == name {
+                for v in v.split(',') {
+                    strings.push(v.trim());
+                }
+            }
+        }
+        strings
+    }
+
     pub fn str(&self, doc_id: &str, line_number: usize, name: &str) -> Result<&str> {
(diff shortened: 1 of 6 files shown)


src/p1/header.rs
```diff
@@ -279,6 +279,21 @@ impl Header {
...
+    pub fn strings(&self, name: &str) -> Vec<&str> {
+        let mut strings = vec![];
+        for (_, k, v) in self.0.iter() {
+            if k.starts_with('/') {
+                continue;
+            }
+            if k == name {
+                for v in v.split(',') {
+                    strings.push(v.trim());
+                }
```
amitu, line 291: Nopes is wrong behaviour based on function name. `.string()` and `.strings()` do completely different things. `.strings()` should not do comma separated behaviour and return all headers matching a given name, and the called then can do a further csv handling.

amitu:
Nope. Not implementing this. Its an anti pattern. Read the [FPM Journal](https://fpm.dev/journal/#the-struggle-with-prelude) page for details.

PR #200, opened 2022-05-02
ftd.type dependencies as class

Arpita-Jaiswal:
…class

### DIFF
--- a/ftd.html
+++ b/ftd.html
@@ -15,6 +15,7 @@
     </script>
     <style>
         __ftd_css__
+        __ftd_element_css__
     </style>
 
 </head>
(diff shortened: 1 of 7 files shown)


ftd.html
```diff
@@ -15,6 +15,7 @@
     </script>
     <style>
         __ftd_css__
+        __ftd_element_css__
```
amitu, line 18: Don't create this, and append `__ftd_css__` and `__ftd_element_css__` in Rust side.

ftd.js
```diff
@@ -1013,10 +1015,24 @@ window.ftd.post_init = function () {
...
+            document.body.classList.add(XL_CLASS);
+            if (document.body.classList.contains(MOBILE_CLASS)) {
+                document.body.classList.remove(MOBILE_CLASS);
+            }
+            return "xl";
+        }
+        if (document.body.classList.contains(MOBILE_CLASS)) {
+            document.body.classList.remove(MOBILE_CLASS);
+        }
+        if (document.body.classList.contains(XL_CLASS)) {
```
amitu, line 1034: The if condition is not really needed, we can unconditionally call the remove function.

PR #220, opened 2022-06-03
Registered and added a SessionDiagnostic

AviralVerma13:
To identify and give error message when a property is declared again (becomes redundant)

### DIFF
--- a/src/p2/interpreter.rs
+++ b/src/p2/interpreter.rs
@@ -56,6 +56,54 @@ impl<'a> Interpreter<'a> {
             self.parsed_libs.push(name.to_string());
         }
     }
+    
+    register_diagnostics! {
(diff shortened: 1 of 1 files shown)


src/p2/interpreter.rs
```diff
@@ -56,6 +56,54 @@ impl<'a> Interpreter<'a> {
...
+
+pub(crate) struct AlreadyDeclared {
+    pub field_name: Ident,
+    #[primary_span]
+    #[label]
+    pub span: Span,
+    #[label = "already-declared-label"]
+    pub prev_span: Span,
+}
+
```
AviralVerma13, line 76: These are the Fluent messages given by the slug = Redundant Property

PR #222, opened 2022-06-08
Used header map to throw error on redundant property

AviralVerma13:
Keep counter of headers with header map and throw error if a counter reaches 2, i.e, a header is repeated/declared again.


AviralVerma13:
The [documentation pull request](https://github.com/FifthTry/ftd.dev/pull/10) has been attached sucessfully.

PR #238, opened 2022-06-23
Duplicate header checking 

Heulitig:
- Added duplicate header checking for components, variables (record)
- Section and subsection level header duplicate checking 
- Duplicate lists are ignored within record variables if those are initialised in header region.
- Invalid header checks for record variables if those are not defined during definition.

## TODO:
- Code refactoring needs to be done.

For variables
- Header Checking for list of records defined inside record variable
- Header Checking for or-type variables defined inside record variable


Heulitig:
This PR is continued in a seperate PR - https://github.com/FifthTry/ftd/pull/338

PR #240, opened 2022-06-23
Commented sections were getting initialized 

Heulitig:
- Commented sections were getting processed inside interpreter
- Ignored those commented sections which was getting processed while reordering the document. 
- Refer this issue - https://github.com/FifthTry/ftd/issues/239


AbrarNitk:
- Skip comment section before interpreting the document.

PR #245, opened 2022-06-28
Ignored comments in interpreter before reordering

Heulitig:
- reverted the reorder() in utils.rs, it doesn't ignore commented sections anymore. 
- commented things are ignored in a separate function before reorder() is called
- after calling this function, the ParsedDocument::sections will only have uncommented portion of the original parsed document.


src/p2/interpreter.rs
```diff
@@ -741,6 +745,102 @@ impl ParsedDocument {
         self.processing_imports = false;
     }
 
+    fn ignore_comments(&mut self) -> ftd::p1::Result<()> {
+        let mut processed_sections: Vec<ftd::p1::Section> = Vec::new();
+        for section in self.sections.iter() {
```
Arpita-Jaiswal, line 750:
```suggestion
        for section in self.sections.iter_mut() {
```

src/p2/interpreter.rs
```diff
@@ -741,6 +745,102 @@ impl ParsedDocument {
...
+
+            let mut filtered_section = ftd::p1::Section {
+                name: section.name.clone(),
+                caption: section.caption.clone(),
+                header: Default::default(),
+                body: None,
+                sub_sections: Default::default(),
+                is_commented: section.is_commented,
+                line_number: section.line_number,
+            };
```
Arpita-Jaiswal, line 764: Use section as `mut` only. `.iter_mut()` would do it. No need to create this variable

Arpita-Jaiswal (changes requested):

Arpita-Jaiswal:
@Heulitig Search for all the places from where the comments are ignored and remove them. Also that places where escape comment is handled, remove them too.

Make sure everything is working fine with your changes

src/p2/interpreter.rs
```diff
@@ -70,9 +70,15 @@ impl InterpreterState {
             }
             self.document_stack[l].done_processing_imports();
 
+            // println!("\nBefore ignoring comments !! ===========================================\n");
+            // self.document_stack[l].show_document();
+
```
amitu, line 75: Don't add these comments, delete them.

src/p2/interpreter.rs
```diff
@@ -745,24 +752,44 @@ impl ParsedDocument {
         self.processing_imports = false;
     }
 
+    /// prints the parsed document (for debugging purposes)
+    fn show_document(&self) {
+        for section in self.sections.iter() {
+            println!(
+                "section name: {}, caption: {:?}, is_commented: {}",
+                &section.name, &section.caption, section.is_commented
+            );
```
amitu, line 762: Use `dbg!()` for debugging, not print. It is less typing and gives more info, and it makes it obvious to everyone that it is a debug statement and not an actual requirement.

Heulitig: Ohh okay

src/p2/interpreter.rs
Heulitig: I was just testing this I will remove these comments and that debug function once all conditions are passed in the final commit.

src/p1/section.rs
```diff
@@ -36,18 +36,7 @@ impl Section {
...
-                _ if body.1.starts_with(r"\/") =>
-                {
-                    #[allow(clippy::single_char_pattern)]
-                    Some((body.0, body.1.strip_prefix(r"\").expect("").to_string()))
-                }
-                _ if body.1.starts_with('/') => None,
-                _ => self.body.clone(),
-            },
-        };
+        let body = self.body_without_comment();
```
Arpita-Jaiswal, line 39: Why is this needed? Is this not already taken care of in `ignore_comments`? If not so, please do it then. This should be just the `body`. if there's any redundant code, please remove them.

src/p1/sub_section.rs
```diff
@@ -59,18 +59,7 @@ impl SubSection {
...
-                _ if body.1.starts_with(r"\/") =>
-                {
-                    #[allow(clippy::single_char_pattern)]
-                    Some((body.0, body.1.strip_prefix(r"\").expect("").to_string()))
-                }
-                _ if body.1.starts_with('/') => None,
-                _ => self.body.clone(),
-            },
-        };
+        let body = self.body_without_comment();
```
Arpita-Jaiswal, line 62: Same as above

src/p2/interpreter.rs
```diff
@@ -45,6 +45,11 @@ impl InterpreterState {
 
         let l = self.document_stack.len() - 1; // Get the top of the stack
 
+        // Any of these two functions can be used to remove comments from parsed document
+        // self.document_stack[l].ignore_all_comments()?;
```
Arpita-Jaiswal, line 49: Choose the best and keep only one

src/p2/interpreter.rs
```diff
@@ -741,6 +737,157 @@ impl ParsedDocument {
...
+            // Filter section body
+            section.body = {
+                match section.body {
+                    Some(ref b) if b.1.trim().is_empty() => None,
+                    Some(ref b) if b.1.trim().starts_with('/') => None,
+                    // To allow '/content' in section body, we need to use "\/content"
+                    // while stripping out the initial '\' from this body
+                    Some(ref b) if b.1.trim().starts_with(r"\/") => {
+                        Some((b.0, b.1.trim().replacen(r"\", "", 1)))
+                    }
```
Arpita-Jaiswal, line 832: Check this for `headers` and `caption` too. Also use one function which does this in all headers, body and caption in sections and subsections.

Arpita-Jaiswal (changes requested):

src/p2/interpreter.rs
```diff
@@ -741,6 +737,157 @@ impl ParsedDocument {
...
+                );
+
+                for (_ln, key, value) in subsection.header.0.iter() {
+                    dbg!(key, value);
+                }
+
+                dbg!(&subsection.body);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
Heulitig: removed

src/p2/interpreter.rs
```diff
@@ -741,6 +737,157 @@ impl ParsedDocument {
...
+    #[allow(dead_code)]
+    fn ignore_all_comments(&mut self) -> ftd::p1::Result<()> {
+        self.sections = self
+            .sections
+            .iter()
+            .filter(|s| !s.is_commented)
+            .map(|s| s.remove_comments())
+            .collect::<Vec<ftd::p1::Section>>();
+        Ok(())
+    }
```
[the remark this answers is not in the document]
Heulitig: This is used now, the other one is removed

src/p1/section.rs
Heulitig: body_without_comment() is only used by remove_comments() which in turn is only used by ignore_comments() in the interpreter.

src/p2/interpreter.rs
Heulitig: I have removed the other one

src/p1/sub_section.rs
Heulitig: Same as above

src/p2/interpreter.rs
```diff
@@ -741,6 +737,157 @@ impl ParsedDocument {
...
+    ///
+    /// 5.  /sub-section-header: value
+    ///
+    /// ## Note: To allow ["/content"] inside body, use ["\\/content"].
+    ///
+    /// Only '/' comments are ignored here.
+    /// ';' comments are ignored inside the [`parser`] itself.
+    ///
+    /// [`parser`]: ftd::p1::parser::parse
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
Heulitig: This just creates a link to that function in the functional doc within the crate in case if anyone wants to refer its definition.

src/p2/interpreter.rs
Heulitig: Done

PR #258, opened 2022-07-18
add `default-value` to `ftd.input`

Heulitig:
- added default-value for ftd.input.
- refer issue - https://github.com/FifthTry/ftd/issues/255

# Usage: 

```ftd
-- ftd.input: 
default-value: Enter text
```

[Doc PR](https://github.com/FifthTry/ftd.dev/pull/21)


src/p2/element.rs
```diff
@@ -1691,7 +1696,16 @@ pub fn input_from_properties(
...
-        value: ftd::p2::utils::string_optional("value", properties, doc.name, 0)?,
+        value: {
+            // if value is unchanged then
+            // value = default-value
+            // otherwise retain the changed value
+            match value {
+                Some(v) => Some(v),
+                None => default_value.clone(),
+            }
+        },
```
amitu, line 1707: This means even if author doesn't want `value` you are still putting `value`. Remember DOM behaviour differs based on if you have provided `value` or `default-value`. We are implementing `default-value` because we do not want to set `value`. 

Why are you doing this?

amitu:
@Heulitig also create a PR in ftd.dev to document this property.

src/p2/element.rs
Heulitig: So you mean if we give default value we need to handle it seperately by generating 
<input type="text", value=default-value, ...> instead of relying on value for generating the same ?

amitu: If there is `default-value`, the HTML will contain `<input defaultValue="default-value">` (or the JS equivalent), not `value=""`.

src/p2/element.rs
```diff
@@ -1696,16 +1694,8 @@ pub fn input_from_properties(
...
-            // value = default-value
-            // otherwise retain the changed value
-            match value {
-                Some(v) => Some(v),
-                None => default_value.clone(),
-            }
-        },
-        defaultvalue: default_value,
+        value: ftd::p2::utils::string_optional("value", properties, doc.name, 0)?,
+        defaultvalue: ftd::p2::utils::string_optional("default-value", properties, doc.name, 0)?,
```
amitu, line 1698: We use camelCase in Rust. `_` is also fine. Just merging two words is not ok.

Heulitig: rust suggests to use snakecase like this default_value

src/p1/mod.rs
```diff
@@ -19,6 +19,9 @@ pub enum Error {
         line_number: usize,
     },
 
+    #[error("forbidden usage: {message}, doc: {doc_id}")]
+    ForbiddenUsage { message: String, doc_id: String },
```
amitu, line 23: Include line number also here, else it would be hard for the author to figure out where the input it. Remember a ftd file may be 100s of lines long with multiple ftd.inputs.

Heulitig: Ohh okay

Heulitig: So the line number has to point the section (--ftd.input) or these headers (value/default-value) ?

examples/ftd-input-default-value.ftd
```diff
@@ -0,0 +1,53 @@
...
+$on-input$: $foo_2=$VALUE
+min-width: 150
+min-height: 50
+color: $white
+background-color: $black
+
+--- ftd.text: Foo2 Text below (changes on input-above)
+padding-bottom: 10
+
+--- ftd.text: $foo_2
```
Arpita-Jaiswal, line 53: Insert a new line

Arpita-Jaiswal (approved):

PR #259, opened 2022-07-21
Code Cleanup

AbrarNitk:


src/p2/document.rs
```diff
@@ -620,7 +620,15 @@ impl Document {
         if k.contains('#') {
             k.to_string()
         } else {
-            format!("{}#{}", self.name.as_str(), k)
+            format!(
+                "{}#{}",
+                if self.name.eq("FPM") {
+                    self.name.to_lowercase()
+                } else {
```
amitu, line 627: Why this?

src/p2/document.rs
```diff
@@ -783,7 +801,11 @@ impl Document {
         match v {
             ftd::PropertyValue::Value { value, .. } => self.value_to_json(value),
             ftd::PropertyValue::Reference { name, .. } => self.json(name),
-            _ => unreachable!(),
+            ftd::PropertyValue::Variable { .. } => Err(ftd::p1::Error::TypeNotCovertible {
+                doc_id: "".to_string(),
+                line_number: 0,
```
amitu, line 806: Pass these two parameters to `object_to_json()`. We always call this method from a processor and processor has section, which contains line number.

AviralVerma13:
cargo test failing for **test p2::record::test::list**

Parse error for **"body is missing in friends"**

will be solved by Rithik's wip PR  "**Reworked caption body checks on components**" https://github.com/FifthTry/ftd/pull/325

PR #328, opened 2022-08-03
added type attribute in input field

priyanka9634:
added type attribute in the input field.


Arpita-Jaiswal (approved):

amitu:
We should have a check on the allowed values of type. [Deferring it for future](https://github.com/FifthTry/ftd/discussions/332).

PR #330, opened 2022-08-04
ignoring escape caption '\/' when ignoring comments

Heulitig:
- ignoring escape caption \/ when ignoring comments.


amitu:
@Heulitig maybe close this PR.

PR #337, opened 2022-08-08
Introducing new BG color as `fpm.color.main.background.code`

gsalunke:
We have introduced new BG color for code block as `#2B303B` which helps all `FTD` themes to use Code-Block with same BG inside any component.

Usage:
```ftd
-- ftd.column:
background-color: $fpm.color.main.background.code`
```


amitu:
Nice! First Rust PR by @gsalunke \o/.

amitu:
Now that I think about it, the [`ftd.color:theme`](https://ftd.dev/code/) should also be part of color scheme, so both dark mode and light mode theme for syntax highlighting can be updated when I use a new color-scheme.

PR #338, opened 2022-08-08
reworked duplicate header check on components

Heulitig:
- This PR is a continuation of an [old PR](https://github.com/FifthTry/ftd/pull/238) 
- issue - https://github.com/FifthTry/ftd/issues/181
- revamped the original PR

# Things added 

- added restriction on defining multiple headers with same name/identifier. 
```markdown
;; This is forbidden now
-- ftd.row foo:
caption name:
string name:
```
- assigning value mulitple times on the same header is restricted as well. Exception being the headers with `list` type
```markdown
;; This is not allowed as well
-- ftd.text: Hello friends
align: center
align: left
```
- added tests inside component.rs
- added a macro `intf!` for `interpreter failing` assertion similar to `f!`

# Note
- Tested these checks on other repos - `dotcom`, `doc-site`, `ftd.dev`, `fpm.dev`, `df`

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -89,6 +89,37 @@ macro_rules! p {
     };
 }
 
+macro_rules! intf {
+    ($s:expr, $m: expr,) => {
+        int_fail!($s, $m)
+    };
+    ($s:expr, $m: expr) => {
+        match ftd::test::interpret("foo", indoc::indoc!($s), &ftd::p2::TestLibrary {}) {
+            Ok(some_value) => panic!("expected failure {:?}, found: {:?}", $m, some_value),
+            Err(e) => {
+                let expected_error = $m.trim();
+                let err_found = e.to_string();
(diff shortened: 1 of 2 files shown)


Arpita-Jaiswal:
@Heulitig is this tested against fpm? and also `fpm.dev`, `ftd.dev`, `dotcom` and `df`?

src/test.rs
```diff
@@ -89,6 +89,37 @@ macro_rules! p {
     };
 }
 
+macro_rules! intf {
```
amitu, line 92: What does `intf` stand for?

Heulitig: It's for `Interpreter failing` assertion. There is a similar macro f! for `parser failing` assertion inside `parser.rs` so I thought of doing something similar.

Arpita-Jaiswal (approved):

PR #339, opened 2022-08-08
conflicting text weight style

ADITIRAI290698:
Added error messages in ui.rs

### DIFF
--- a/src/p1/header.rs
+++ b/src/p1/header.rs
@@ -36,40 +36,6 @@ impl Header {
             .push((*line_number, name.to_string(), value.to_string()))
     }
 
-    // Returns error is text style is assigned two or more weights
-    pub fn conflict_style(&self) -> ftd::p1::Result<()> {
-        let mut hm = std::collections::HashMap::new();
-        for (ln, k, v) in self.0.iter() {
-            if k.starts_with("style") {
-                let mut count = 0;
-                for i in [
-                    "heavy",
-                    "extra-bold",
-                    "bold",
-                    "semi-bold",
-                    "medium",
-                    "regular",
-                    "light",
-                    "extra-light",
-                    "hairline"
-                ] {
-                    if v.contains(i) {
-                        count += 1;
-                        hm.insert(v, i);
-                        if count >= 2 {
-                            return Err(ftd::p1::Error::ParseError {
-                                doc_id: k.to_string(),
-                                line_number: *ln,
-                                message: "conflicting text weight styles found".to_string(),
-                            });
-                        }
-                    }
-                }
-            }
-        }
-        Ok(())
-    }
-
     pub fn bool_with_default(
         &self,
         doc_id: &str,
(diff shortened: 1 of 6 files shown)


src/p1/header.rs
```diff
@@ -36,6 +36,40 @@ impl Header {
             .push((*line_number, name.to_string(), value.to_string()))
     }
 
+    // Returns error is text style is assigned two or more weights
+    pub fn conflict_style(&self) -> ftd::p1::Result<()> {
```
amitu, line 40: This logic should be inside the `ftd::Style::from()` method.

amitu:
@ADITIRAI290698 This logic should be inside the ftd::Style::from() method.

src/ui.rs
```diff
@@ -2656,21 +2656,23 @@ impl Style {
...
+                "heavy" if s.weight.is_none() => s.weight = Some(ftd::Weight::Heavy),
+                "extra-bold" if s.weight.is_none() => s.weight = Some(ftd::Weight::ExtraBold),
+                "bold" if s.weight.is_none() => s.weight = Some(ftd::Weight::Bold),
+                "semi-bold" if s.weight.is_none() => s.weight = Some(ftd::Weight::SemiBold),
+                "medium" if s.weight.is_none() => s.weight = Some(ftd::Weight::Medium),
+                "regular" if s.weight.is_none() => s.weight = Some(ftd::Weight::Regular),
+                "light" if s.weight.is_none() => s.weight = Some(ftd::Weight::Light),
+                "extra-light" if s.weight.is_none() => s.weight = Some(ftd::Weight::ExtraLight),
+                "hairline" if s.weight.is_none() => s.weight = Some(ftd::Weight::HairLine),
+                _t => return ftd::e2(format!("\'{}\' is not a valid style", &l), doc_id, 0),
```
Arpita-Jaiswal, line 2675:
```suggestion
                _ => return ftd::e2(format!("\'{}\' is not a valid style", &l), doc_id, 0),
```

amitu: Also I am not happy with such generic error message. Let's give more precise error message. We want to write more code so that end user has to use less brain to figure out what exactly is wrong.

src/p2/element.rs
```diff
@@ -1163,6 +1163,7 @@ pub fn color_from(l: Option<String>, doc_id: &str) -> ftd::p1::Result<Option<ftd
             return ftd::e2(format!("{} is not a valid color", v), doc_id, 0);
         }
 
+        //Code for accepting 6-digit hexa-color code
```
amitu, line 1166: I think this comment is wrong. Look at the if condition this code is part of.

src/ui.rs
```diff
@@ -2655,24 +2663,96 @@ impl Style {
...
             }
         }
+
+        // Checks if there is repeatation in Underline,italic,strike
+        for (style, count) in booleans.iter() {
+            if count > &1 {
+                return Err(ftd::p1::Error::ForbiddenUsage {
+                    message: format!("\'{}\' repeated {} times in \'{}\'", style, count, &l),
+                    doc_id: doc_id.to_string(),
+                    line_number: 0,
```
amitu, line 2731: The line number is wrong. Pass line number from the caller (of the from method) if need be.

PR #354, opened 2022-08-17
adding StuckOnTerm interpreter state for processing terms

Heulitig:
- working on StuckOnTerm state for processing terms and to replace those with links. **(wip)** 

## TODO: 

This will be done in a seperate PR **(wip)**
- need to add `term` as `universal argument` 
- need to segregate `common` kernel arguments and `universal` component arguments (only in doc)
- need to refactor `common` arguments to `common kernel` arguments (only in doc)

### DIFF
--- a/src/p2/interpreter.rs
+++ b/src/p2/interpreter.rs
@@ -1,5 +1,3 @@
-use std::borrow::Borrow;
-
 #[derive(Debug, Default)]
 pub struct InterpreterState {
     pub id: String,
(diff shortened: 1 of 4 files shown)


src/p2/interpreter.rs
```diff
@@ -741,6 +766,38 @@ impl ParsedDocument {
...
+        _terms_map: &std::collections::HashMap<String, String>,
+    ) -> ftd::p1::Result<()> {
+        // Term syntax 1 = [term`](term: someTerm)
+        // Refer someTerm from linked text <term`>
+        // replacement = \[[term`]\]([document-id]#[slugified(someTerm)])
+        let _pattern_1 = regex::Regex::new(
+            r"(?x) # Enabling Comment Mode
+            \[(?P<linked_text>[\sa-zA-Z\d]+)\] # Linked Text Capture Group <linked_text>
+            \(\s*term\s*:(?P<actual_term>[\sa-zA-Z\d]+)\) # Referred Term Capture Group <actual_term>"
+        ).unwrap();
```
amitu, line 784: You can compile regexes and store as lazy static so you can cache compilation output. 

Checkout: https://docs.rs/regex/1.6.0/regex/#example-avoid-compiling-the-same-regex-in-a-loop

amitu: Do a search in entire code base where we have used regex without lazy static, and use this recommendation.

Heulitig:
This PR is outdated. Now instead of StuckOnTerm state there will be a CheckID state in interpreter which will check for unique ids and link based on those ids

PR #361, opened 2022-08-18
refactoring common properties to common kernel properties

Heulitig:
- just a naming refactor to change `common` properties to `common kernel` properties **(wip)**
- these `common kernel` properties are available on every `kernel` component


Heulitig:
No need for this code refactor closing it

PR #374, opened 2022-08-24
added ftd::Set<T> type definition

Heulitig:
- [x] added type definition `ftd::Set<T>` for `std::collections::HashSet<T>` in lib.rs
- [x] refactored the original usage of `std::collections::HashSet` with `ftd::Set` throughout the code.

### DIFF
--- a/src/code.rs
+++ b/src/code.rs
@@ -13,7 +13,7 @@ lazy_static::lazy_static! {
         }
         builder.build()
     };
-    pub static ref KNOWN_EXTENSIONS: std::collections::HashSet<String> =
+    pub static ref KNOWN_EXTENSIONS: ftd::Set<String> =
         SS.syntaxes().iter().flat_map(|v| v.file_extensions.to_vec()).collect();
     pub static ref TS: syntect::highlighting::ThemeSet =
         syntect::highlighting::ThemeSet::load_defaults();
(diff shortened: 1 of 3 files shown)


amitu:
Nopes. What is the need for this? We created `fpm::Map` because we want to be able to switch between `HashMap` and `BTreeMap` and we one in test and other in production. Why are we creating `fpm::Map`?

Heulitig:
> Nopes. What is the need for this? We created `fpm::Map` because we want to be able to switch between `HashMap` and `BTreeMap` and we one in test and other in production. Why are we creating `fpm::Map`?

But I'm creating `ftd::Set<T>` not `fpm::Map<T>`. Well since you are insisting, I will close this PR then.

amitu:
> But I'm creating `ftd::Set<T>` not `fpm::Map<T>`.

I understand. I was saying we created `fpm::Map<T>` for so and so reason. What is the reason for creating `ftd::Set<T>` alias?

> Well since you are insisting, I will close this PR then.

I am not insisting. I am giving reason for my action. I am asking reason for yours.

amitu:
Also this was my bad, I meant "Why are we creating `fpm::Set`" not "Why are we creating `fpm::Map`". I think this caused the confusion.

PR #402, opened 2022-09-02
markdown color changes

Heulitig:
- markdown colors updated


Heulitig:
@Arpita-Jaiswal 
- some style colors are getting duplicated after updating the markdown colors using update_markdown_colors() in ftd.js 
- For now, the texts are visibly clear after these changes

PR #423, opened 2022-09-15
regex adjustments for change in linking syntax

Heulitig:
Discussion: [global id](https://github.com/FifthTry/fpm/discussions/450)
- wip


Arpita-Jaiswal (changes requested):

src/p2/utils.rs
```diff
@@ -1561,3 +1588,67 @@ where
...
+pub fn get_thing_ignore_fail(
+    name: &str,
+    line_number: usize,
+    doc: &ftd::p2::TDoc,
+) -> Option<ftd::p2::Thing> {
+    match doc.get_thing(line_number, name) {
+        Ok(thing) => Some(thing),
+        Err(_) => None,
+    }
+}
```
Arpita-Jaiswal, line 1625:
```suggestion
doc.get_thing(line_number,  name).ok()
```

You don't need `get_thing_ignore_fail` function at all. Use `.ok()` instead

Arpita-Jaiswal (changes requested):

src/p2/utils.rs
Arpita-Jaiswal: @Heulitig FYI

PR #448, opened 2022-10-13
Auto numbering of headings

Heulitig:
[FPM PR](https://github.com/ftd-lang/fpm/pull/560)
- auto-numbering of headings
- adjusted to work with headings with auto-generated-ids
- other fixes


src/ui.rs
```diff
@@ -2329,6 +2329,7 @@ pub struct Common {
     pub border_style: Option<String>,
     pub text_transform: Option<String>,
     pub title: Option<String>,
+    pub h_number: Option<String>,
```
amitu, line 2332: Rename this to `heading_number`. Move it to TextCommon or something. And the type should be `Vec<String>` instead of `Option<String>`. So one can pass `["1", "a", "i"]` and the doc-site can decide to show dots or dashes etc.

PR #464, opened 2022-11-21
bury implementation in ftd

AviralVerma13:


src/lib.rs
```diff
@@ -115,6 +115,7 @@ pub struct PageHeadingItemCompat {
     pub number: Option<String>,
     pub title: Option<String>,
     pub path: Option<String>,
+    pub bury: bool,
```
amitu, line 118: Why do we need this field in "page-headings processor" struct? This is a sitemap feature.

AviralVerma13: okay. got it. closing the PR now.

AviralVerma13:
not needed here.

PR #522, opened 2023-01-09
fixed issue with literals having hyphens

Heulitig:
- literals with hyphens were treated as math expressions with minus
- fixed this above issue


Heulitig:
Merging this for now will fix the tests later

PR #537, opened 2023-01-18
ftd handling cache for ParsedDocument

AbrarNitk:


src/p2/mod.rs
```diff
@@ -17,3 +17,9 @@ pub use kind::Kind;
 pub use library::TestLibrary;
 pub use record::Record;
 pub use tdoc::TDoc;
+
+/// ftd is providing this trait to hosts to implement cache for the ParsedDocument.
+pub trait Cache {
+    fn get(id: &str) -> ftd::interpreter2::Result<Option<ftd::ParsedDocument>>;
+    fn set(id: &str, document: &ftd::ParsedDocument) -> ftd::interpreter2::Result<()>;
+}
```
amitu, line 25: Don't do this, at the right place use a HashMap.

amitu: Cache has to be persisted across requests, so the `&mut HashMap` has to be passed from FPM to ftd in `ftd::interpret()` call.

FPM will store the Hashmap in a global place.

Since we are going to then acquire the lock to access Hashmap, its better to pass the `std::sync::RwLock<HashMap>` to `ftd::interpret()` so FTD acquires the right lock to do the right operation. FTD will first do a read operation to see if it's already parsed by taking read lock. If it is not in the Hashmat, it will take write lock, parse and update the Hashmap.

AbrarNitk:
Closing this PR because we have to implement it with different approach.

PR #651, opened 2023-03-10
Added Shorthand list + border style updated

Heulitig:


amitu:
@Heulitig I have changed a couple of `Option<Vec<T>>` to `Vec<T>`s. Usually it does not make sense to wrap a list in optional, as often null case and empty list cases are the same.

PR #765, opened 2023-03-21
`ftd.document`

Arpita-Jaiswal:


ftd/src/executor/element.rs
```diff
@@ -83,6 +111,18 @@ pub struct Column {
     pub common: Common,
 }
 
+#[derive(serde::Deserialize, Debug, Default, PartialEq, Clone, serde::Serialize)]
+pub struct HTMLData {
+    pub title: ftd::executor::Value<Option<String>>,
+}
```
amitu, line 117: @Arpita-Jaiswal: Do we need this type? I think we should merge it with `Document`. I do not want to mention `HTML` in our core type hierarchy as we are supposed to agnostic of rendering layer.

ftd/t/html/86-ftd-document.ftd
```diff
@@ -0,0 +1,7 @@
+-- ftd.document:
+title: My title
```
amitu, line 2: Can we make `title` a caption?

PR #766, opened 2023-03-21
adding background image

Heulitig:


ftd/src/executor/styles.rs
```diff
@@ -1,3 +1,5 @@
+use ftd::executor::ImageSrc;
```
Arpita-Jaiswal, line 1: Remove the `use` statement

ftd/src/node/main.rs
```diff
@@ -1140,12 +1140,25 @@ impl ftd::executor::Common {
             ),
         );
 
+        d.check_and_insert(
```
Arpita-Jaiswal, line 1143: Place todo here. This doesn't solve the conditionality yet.

Arpita-Jaiswal (changes requested):

PR #771, opened 2023-03-23
added background repeat css

Heulitig:


Heulitig:
This property will be included under background-image record, so this is no longer needed.

PR #826, opened 2023-04-05
`fastn` is same as `fastn serve`

amitu:
We will do this when the [next edition](https://fastn.io/d/next-edition/) goes live.


amitu:
Re-doing this directly in main.

PR #1023, opened 2023-07-06
Fixed Semicolon Bug and Added String Support in `extract_arguments` function

theharshsingh:
Hey Team

In the update code I have fixed the semicolon bug where `SELECT * FROM users WHERE user = $name;` won't be parsed correctly because there was no whitespace after it. In this update, insted of checking for next whitespace, we are checking for next special characters like `':', ";", '"',` etc.

Next thing I added is support for strings. It supports both single and double quote strings, checks if it is closed or not, and won't count anything inside a string as a param eg. `SELECT * FROM users WHERE user = "$name";` -- this will not be counted as a param, but will be treated as a regular string, it also supports escape sequence. One thing it lacks is throwing an error if the string is left open, although we are keeping track of it and I have implemented the code for it, I was not sure which error we should throw.

Lastly, the code performance can be improved a lot. Maybe we should use RegEx? Please give your suggestions.

### DIFF
--- a/.env
+++ b/.env
@@ -1,1 +1,1 @@
-export FASTN_PG_URL="postgres://postgres:[YOUR-PASSWORD]@db.impmoeegmmnnewsdwdwk.supabase.co:6543/postgres"
+export FASTN_PG_URL="postgres://postgres:12deep45@localhost/harsh"
--- a/fastn-core/src/library2022/processor/sql.rs
+++ b/fastn-core/src/library2022/processor/sql.rs
@@ -31,7 +31,7 @@ pub(crate) fn extract_arguments(query: &str) -> ftd::interpreter::Result<(String
             }
         }
 
-        if chars[i] == '"' || chars[i] == '"' && !escaped {
+        if (chars[i] == '"' || chars[i] == '"') && !escaped {
             if quote_open {
                 if Some(chars[i]) == quote {
                     quote_open = false;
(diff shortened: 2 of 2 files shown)


amitu:
@theharshsingh, please remove the `#[ignore]` from the test and add more test for every feature you have implemented.

theharshsingh:
> @theharshsingh, please remove the `#[ignore]` from the test and add more test for every feature you have implemented.

Ok

amitu:
Also let's delete the `.env` file, and add a `.gitignore` for it.

amitu:
@theharshsingh regarding performance, I wouldn't worry about it for now. ftd itself is slow, and we have database access, which is a over the network call, I would be really surprised if this function contributes even 1% time to end to end request-response duration. We should solve the ftd parser/interpreter performance first.

theharshsingh:
> Also let's delete the `.env` file, and add a `.gitignore` for it.

I also thought of adding a .gitignore



> @theharshsingh regarding performance, I wouldn't worry about it for now. ftd itself is slow, and we have database access, which is a over the network call, I would be really surprised if this function contributes even 1% time to end to end request-response duration. We should solve the ftd parser/interpreter performance first.

Ok. I will try to improve it later on, and yes, FTD parser can be made much faster.

theharshsingh:
Fixed the quote bug and added tests for testing the same. Now its working fine.

amitu:
@theharshsingh the build is failing, you can click on the Red Cross and check the message and fix them.

PR #1053, opened 2023-07-13
Moved images from root folder to asset folder

abhiramiyer:


amitu:
@abhiramiyer since these files are only needed on Github, maybe we can move them to `.github` folder.

Pro: one less folder on top level. Con: Github may not like random stuff in `.github` someday. What do you think?

Arpita-Jaiswal (approved):

abhiramiyer:
@amitu These files are needed even for local rendering e.g. when I open the README.md in any Markdown renderer or VSCode. Hence I don't think these should go in .github folder. There is an alternative way to avoid creating another top level folder for this particular scenario by using a dedicated 'asset' branch instead: https://gist.github.com/joncardasis/e6494afd538a400722545163eb2e1fa5. In this approach, instead of adding images to the asset folder, we add them to the 'asset' branch. Does this look better?

amitu:
Let's go ahead with this for now.

Issue #958, opened 2023-06-13
Feature: Inline comments


Arpita-Jaiswal:
Currently, we add comments using `;;` at the beginning of the line in fastn language.
We need to extend this feature to be used inline.

### Case 1
```ftd
-- ftd.text: Hello      ;; displays the hello
```
The Output should be:
```
Hello
```
This should also be working.

### Case 2
We can escape the comments inline by using `\`.

```ftd
-- ftd.text: Hello      \;; displays the hello
```
The Output should be:
```
Hello      ;; displays the hello
```

### Case 3
We can also escape `\`.
```ftd
-- ftd.text: Hello      \\;; displays the hello
```
The Output should be:
```
Hello      \;; displays the hello
```

harshdoesdev:
So, there are two ways to start a comment:
/-- for commenting out sections
; for commenting a single line or part of line (not implemented yet)

I found that the /-- comments are being ignored by the interpreter, while the ; are ignored by the parser itself. ; can be escaped using \ if we want to use that character. So we have to check if a line contains a ;, if yes, then we will have to check if it is not preceded by any \, if it is preceded by a backslash, we will count it as a normal text, and then check again if the string contains any other ;, if yes then repeat the loop, if not then the string does not have any inline comments. But if it was not preceded by any \, then we will split at this index and return the left hand-side string.

Is it a nice approach to this problem?

harshdoesdev:
This issue can be closed because I have implemented the inline comments feature in this PR. Currently, we have used ;; in many places as a marker for highlighting HTML (Refer: HTML test 52) in FTD, which is why some tests are failing. We would need to find a new syntax for that marker because they are syntactically the same and cannot be distinguished without some expensive operations. They have been extensively used in fastn.com, so we need to replace them there first.

Here is the PR:
https://github.com/fastn-stack/fastn/pull/1077

PR #1077, opened 2023-07-17
Added inline commenting 

harshdoesdev:
Fixed the remove_inline_comments function:
- Fixed the escape sequence bug which was preventing the `;` from being escaped when using `\\` double-backslashes
- Changed the regular while loop with while..some while loop
- Fixed formatting issues


ftd/src/ftd2021/p1/parser.rs
```diff
@@ -401,6 +408,71 @@ pub fn parse_file_for_global_ids(data: &str) -> Vec<(String, usize)> {
...
+        };
+        ($s:expr, $t: expr) => {
+            assert_eq!(super::remove_inline_comments($s), $t)
+        };
+    }
+
+    #[test]
+    fn test_esc() {
+        t!(r#"Hello      ;; displays the hello"#, "Hello");
+        t!(r#"Hello      \;; displays the hello"#, "Hello      ;");
```
amitu, line 471: This is wrong. There is no comment in this line as comment requires two consecutive unescaped `;;`.

harshdoesdev:
More updates about why some tests are failing:
https://github.com/fastn-stack/fastn/issues/958#issuecomment-1641429175

amitu:
> Currently, we have used ;; in many places as a marker for highlighting HTML (Refer: HTML test 52) in FTD, which is why some tests are failing. 

We have to change the syntax for code highlighting.

amitu:
For code highlighting we have to put `# <hl>` etc. We will honour line comment based on lang attribute. For each lang lets create a map for line comment. Or we just assume that virtually everyone either uses `#` or `//` as line comment, so we look for `# <hl>` and `// <hl>`, in any other language, people can use either of the two.

ftd/src/p1/parser.rs
```diff
@@ -872,9 +872,70 @@ fn clean_line(line: &str) -> String {
     if line.starts_with("\\;;") || line.starts_with("\\-- ") {
         return line[1..].to_string();
     }
+
+    if !line.contains(" ;; <hl>") {
```
Heulitig, line 876: Just change it to this

```suggestion
    // Ignoring <hl> to ignore code highlighting comments
    // todo: need to fix the code highlighting syntax
    if !line.contains("<hl>") {
```

ftd/src/p1/parser.rs
```diff
@@ -872,9 +872,70 @@ fn clean_line(line: &str) -> String {
...
     line.to_string()
 }
 
+fn remove_inline_comments(line: &str) -> String {
+    let mut output = String::new();
+
+    let mut chars = line.chars().peekable();
+
+    let mut escape = false;
+
```
Heulitig, line 889:
```suggestion
    let mut output = String::new();
    let mut chars = line.chars().peekable();
    let mut escape = false;
    let mut count = 0;
```

Issue #1121, opened 2023-07-26
Error in `install.sh` script for MacOS


gargajit:
The URL is:
```
https://fastn.com/install.sh
```

If the following folder does not exist then this installer is not working.
```
/usr/local/bin
```

amitu:
[Someone is saying you can just create this folder if it is missing](https://stackoverflow.com/questions/25654731/usr-local-bin-no-such-file-or-directory). @harshdoesdev once you are done with you two PRs can you take a stab at this, should be a trivial fix. 

Also if put a check, if you fail to create the folder, show a proper message saying /usr/local/bin not found and can not create also, please help or something.

harshdoesdev:
@amitu @gargajit, this issue is now resolved.

PR: https://github.com/fastn-stack/fastn/pull/1127

amitu:
@gargajit please verify and close this issue.

gargajit:
@harshdoesdev Path is not updating.

Steps followed:
1. Deleted the existing `bin` which had `fastn` from the path `/usr/local/bin/`
2. Downloaded the updated `install.sh` script file and replaced the old file with the updated file in `fastn.com`.
3. Copied the installer script from `fastn.com/install`: `curl -fsSL https://fastn.com/install.sh | bash`
4. In terminal, executed the script (successfully installed)
<img width="573" alt="Screenshot 2023-07-31 at 1 57 42 PM" src="https://github.com/fastn-stack/fastn/assets/118595104/c84e37bc-3068-43d2-bb61-9fbd5d460a59">

5. Quit Terminal, opened new Terminal window, ran the command `where fastn`, Output: `fastn not found`
<img width="845" alt="Screenshot 2023-07-31 at 2 00 18 PM" src="https://github.com/fastn-stack/fastn/assets/118595104/0d91d324-fa20-4407-b2e9-40815fee8e75">

Note: Tried both `bash` and `sh`. No luck.

Checked
- `/usr/local/bin/`
- `/usr/bin/`
- `/Users/ajitgarg/`

amitu:
@gargajit you reported "If the following folder does not exist then this installer is not working: /usr/local/bin". Did you verify that this issue is fixed, that installer works if the `/usr/local/bin` is missing?

gargajit:
> @gargajit you reported "If the following folder does not exist then this installer is not working: /usr/local/bin". Did you verify that this issue is fixed, that installer works if the `/usr/local/bin` is missing?

[...]

PR #1127, opened 2023-07-28
Installer.sh fix and improvements

harshdoesdev:
+ Fixed the `installer.sh` file so that it works even without a '/usr/local/bin' (use $HOME/.fastn/bin dir instead)
+ Also improved the UI and outputs

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -20,25 +20,43 @@ else
   }
 fi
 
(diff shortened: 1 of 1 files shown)


harshdoesdev:
Issue: https://github.com/fastn-stack/fastn/issues/1121

PR #1133, opened 2023-08-01
install.sh fixes

harshdoesdev:
Fixes the below mentioned issues in the `install.sh` file:
- Removes any previously downloaded temporary binary files

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -77,16 +77,6 @@ update_path() {
     fi
 }
 
-remove_existing_files() {
-    local existing_files=(fastn_macos_x86_64 fastn_linux_musl_x86_64 fastn_controller_linux_musl_x86_64)
-
-    for file in "${existing_files[@]}"; do
-        if [ -f "$file" ]; then
-            rm -f "$file"
-        fi
-    done
-}
-
 command_exists() {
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -77,6 +77,16 @@ update_path() {
     fi
 }
 
+remove_existing_files() {
+    local existing_files=(fastn_macos_x86_64 fastn_linux_musl_x86_64 fastn_controller_linux_musl_x86_64)
+
+    for file in "${existing_files[@]}"; do
+        if [ -f "$file" ]; then
+            rm -f "$file"
```
amitu, line 85: Isn't `rm fastn_macos_x86_64 fastn_linux_musl_x86_64 fastn_controller_linux_musl_x86_64` enough?

amitu: You are trying to ignore the error message?

harshdoesdev: yes, maybe the installation failed because insufficient permissions, and the user runs the installer again

harshdoesdev: thanks, changed it to: `rm -f fastn_macos_x86_64 fastn_linux_musl_x86_64 fastn_controller_linux_musl_x86_64`

Heulitig (approved):

PR #1163, opened 2023-08-09
Basic implementation of `ftd.local_storage`

harshdoesdev:
`ftd.local_storage` is a wrapper over the `localStorage` API provided by the browser.

Usage:
```ftd
-- record stuff:
string something:

-- stuff my-stuff:
something: good

-- ftd.text: Click me
$on-click$: $save-stuff(s = $my-stuff)

-- void save-stuff(s):
stuff s:

ftd.local_storage.set("test", s)

-- ftd.text: Get Stuff
$on-click$: $get-stuff(key = test)

-- stuff get-stuff(key):
string key:

console.log(ftd.local_storage.get(key))
```


amitu:
We also need name spacing. Local store data by one fastn package should not be readable by another fastn package. 

@Arpita-Jaiswal how can we get the name of the package that made the call? Eg if `package_a/index.ftd` calls `package_b/utils.do_something`, and `do_something` calls `ftd.local_store.get()`, we only return what `package_b` has access to. We do this by prefixing the keys with `package_b__<key>` etc.

harshdoesdev:
> We also need name spacing. Local store data by one fastn package should not be readable by another fastn package.
> 
> @Arpita-Jaiswal how can we get the name of the package that made the call? Eg if `package_a/index.ftd` calls `package_b/utils.do_something`, and `do_something` calls `ftd.local_store.get()`, we only return what `package_b` has access to. We do this by prefixing the keys with `package_b__<key>` etc.

@amitu The name spacing feature has been added.

amitu:
The implementation is not correct. You are using a single package name for entire document. The package that defined the function that called the local store should be used for name spacing.

harshdoesdev:
> The implementation is not correct. You are using a single package name for entire document. The package that defined the function that called the local store should be used for name spacing.

@amitu This has been fixed and now the correct package name is being used. For example: if a package named `test` contains the utility function `save-data()`, and it is being called from another package by importing it, the package name `test_` will be prefixed when saving data in localstorage.

PR #1183, opened 2023-08-11
serialising `pg` processor queries (to avoid connection/thread issue)

harshdoesdev:
**Fixes:**

The problem was in the `execute_query` function. `fastn` sometimes makes multiple concurrent requests, which leads to attempt of creating multiple statements. since we are using a pool, where statements are pooled so that they can be reused later, it leads to errors like "statement s0 already exists".

We prevent this by using utilizing Mutex Lock, and HashSet, to keep track of already prepared statements and apply a lock when preparing a statement so that only one statement is prepared for these concurrent requests. It then release the lock once it is done.

@amitu

### DIFF
--- a/fastn-core/src/library2022/processor/pg.rs
+++ b/fastn-core/src/library2022/processor/pg.rs
@@ -26,9 +26,10 @@ async fn create_pool() -> Result<deadpool_postgres::Pool, deadpool_postgres::Cre
 static POOL_RESULT: tokio::sync::OnceCell<
     Result<deadpool_postgres::Pool, deadpool_postgres::CreatePoolError>,
 > = tokio::sync::OnceCell::const_new();
+
 static PREPARED_STATEMENT: once_cell::sync::Lazy<
-    tokio::sync::Mutex<std::collections::HashSet<String>>,
-> = once_cell::sync::Lazy::new(|| tokio::sync::Mutex::new(std::collections::HashSet::new()));
+    tokio::sync::RwLock<std::collections::HashSet<String>>,
+> = once_cell::sync::Lazy::new(|| tokio::sync::RwLock::new(std::collections::HashSet::new()));
 
 async fn pool() -> &'static Result<deadpool_postgres::Pool, deadpool_postgres::CreatePoolError> {
     POOL_RESULT.get_or_init(create_pool).await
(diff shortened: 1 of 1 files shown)


fastn-core/src/library2022/processor/pg.rs
```diff
@@ -266,9 +269,19 @@ async fn execute_query(
...
-
+    let stmt_key = query.to_string();
     let client = pool().await.as_ref().unwrap().get().await.unwrap();
-    let stmt = client.prepare_cached(query.as_str()).await.unwrap();
+
+    // Use a Mutex to ensure only one thread prepares the same statement
+    let mut prepared_statements = PREPARED_STATEMENT.lock().await;
+    let stmt = if !prepared_statements.contains(&stmt_key) {
+        let stmt = client.prepare_cached(query.as_str()).await.unwrap();
+        prepared_statements.insert(stmt_key);
```
amitu, line 279: Clone here, instead of line 272. This was most of the time we will not have to clone.

fastn-core/src/library2022/processor/pg.rs
```diff
@@ -266,9 +269,19 @@ async fn execute_query(
...
 ) -> ftd::interpreter::Result<Vec<Vec<serde_json::Value>>> {
     let (query, query_args) = super::sql::extract_arguments(query)?;
-
+    let stmt_key = query.to_string();
     let client = pool().await.as_ref().unwrap().get().await.unwrap();
-    let stmt = client.prepare_cached(query.as_str()).await.unwrap();
+
+    // Use a Mutex to ensure only one thread prepares the same statement
+    let mut prepared_statements = PREPARED_STATEMENT.lock().await;
+    let stmt = if !prepared_statements.contains(&stmt_key) {
```
amitu, line 277: I am not sure about this. The statement is prepared per connection. 

Also if you are using an `if` statement with both `then` and `else` clauses, no need to invert, that forces cognitive load, swap the content of `then` and `else` blocks and remove the `!`.

amitu:
@harshdoesdev let me know when you have tested it and I will merge this.

harshdoesdev:
> @harshdoesdev let me know when you have tested it and I will merge this.

It is working fine on Heroku, though speed is still an issue, but it is not crashing.

PR #1250, opened 2023-08-30
added `ftd.video` element

harshdoesdev:
Added the `ftd.video` element in fastn, which is similar to the HTML5 Video Element.

Example:
```ftd
-- import: video-player/assets

-- ftd.video:
src: $assets.files.videos.flower.webm
controls: true
autoplay: true
muted: true
```

### DIFF
--- a/ftd/src/interpreter/things/default.rs
+++ b/ftd/src/interpreter/things/default.rs
@@ -9938,18 +9938,6 @@ pub fn rive_function() -> ftd::interpreter::ComponentDefinition {
                     access_modifier: Default::default(),
                     line_number: 0,
                 },
-                ftd::interpreter::Argument {
-                    name: "muted".to_string(),
-                    kind: ftd::interpreter::Kind::boolean().into_kind_data(),
-                    mutable: false,
-                    value: Some(ftd::interpreter::PropertyValue::Value {
-                        value: ftd::interpreter::Value::Boolean { value: true },
-                        is_mutable: false,
-                        line_number: 0,
-                    }),
-                    access_modifier: Default::default(),
-                    line_number: 0,
-                },
                 ftd::interpreter::Argument::default(
                     "artboard",
                     ftd::interpreter::Kind::string()


fastn-js/js/dom.js
```diff
@@ -739,7 +746,9 @@ class Node2 {
     }
     // for attaching inline attributes
     attachAttribute(property, value) {
-        if (fastn_utils.isNull(value)) {
+        // If the value is null, undefined, or false, the attribute will be removed.
+        // For example, if attributes like checked, muted, or autoplay have been assigned a "false" value.
+        if (fastn_utils.isNull(value) || value === false) {
```
Heulitig, line 751: Remove this logic from here and add it on individual attributes where it is needed. It will hinder the logic for those attributes where value `false` is used. 
```suggestion
        if (fastn_utils.isNull(value)) {
```

fastn-js/js/dom.js
```diff
@@ -748,6 +757,13 @@ class Node2 {
     removeAttribute(property) {
         this.#node.removeAttribute(property);
     }
+    toggleAttribute(property) {
```
Heulitig, line 760: Remove this function (not used)

ftd/src/interpreter/things/default.rs
```diff
@@ -9841,6 +9938,18 @@ pub fn rive_function() -> ftd::interpreter::ComponentDefinition {
                     access_modifier: Default::default(),
                     line_number: 0,
                 },
+                ftd::interpreter::Argument {
```
Heulitig, line 9941: Remove this `muted` attribute from here and add it on video element

ftd/src/js/element.rs
```diff
@@ -699,6 +704,165 @@ impl Image {
...
+        doc: &ftd::interpreter::TDoc,
+        rdata: &mut ftd::js::ResolverData,
+        should_return: bool,
+    ) -> Vec<fastn_js::ComponentStatement> {
+        let mut component_statements = vec![];
+        let kernel = create_element(fastn_js::ElementKind::Video, parent, index, rdata);
+        component_statements.push(fastn_js::ComponentStatement::CreateKernel(kernel.clone()));
+        component_statements.push(fastn_js::ComponentStatement::SetProperty(
+            fastn_js::SetProperty {
+                kind: fastn_js::PropertyKind::ImageSrc,
```
Heulitig, line 785: Change this to VideoSrc 
```suggestion
                kind: fastn_js::PropertyKind::VideoSrc,
```

amitu:
@harshdoesdev please update docs.

PR #1282, opened 2023-09-08
fastn.redirects syntax update 

Heulitig:


fastn-core/src/package/redirects.rs
```diff
@@ -9,13 +9,20 @@ impl RedirectsTemp {
         let body = self.body.as_str();
         let mut redirects: ftd::Map<String> = ftd::Map::new();
         for line in body.lines() {
-            if line.trim_start().starts_with(';') {
+            if line.is_empty() || line.trim_start().starts_with(';') {
                 continue;
             }
+            // Supported Redirects Syntax under fastn.redirects
+            // <some link>: <link to redirect>
```
amitu, line 16: Let's deprecate this syntax. If you find someone use `:` as separator in `fastn.redirects`, give a warning to them and ask to upgrade to `->` syntax.

PR #1289, opened 2023-09-08
redirect warning fix for old syntax

Heulitig:


fastn-core/src/package/redirects.rs
```diff
@@ -16,17 +16,18 @@ impl RedirectsTemp {
...
             // <some link> -> <link to redirect>
 
+            if let Some((key, value)) = line.split_once("->") {
+                Self::assert_and_insert_redirect(key, value, &mut redirects)?;
+                continue;
+            }
+
             if let Some((key, value)) = line.split_once(':') {
                 fastn_core::warning!(
                     "Redirect syntax: '{key}: {value}' will be deprecated\nPlease use the '{key} \
```
amitu, line 26: @Heulitig this should "Redirect syntax: '{key}: {value}' is derecated..". Deprecated means "this is no longer advisable to use this, it will be removed in future".

PR #1291, opened 2023-09-08
Fix the use of if conditions on `ftd.redirect`

harshdoesdev:
Example:
```ftd
-- record user:
string name:
boolean is_admin:

-- user harsh:
name: Harsh Singh
is_admin: false

-- ftd.redirect: /admin
if: { harsh.is_admin }

-- ftd.redirect: /user 
```

### DIFF
--- a/ftd/src/interpreter/main.rs
+++ b/ftd/src/interpreter/main.rs
@@ -1104,27 +1104,32 @@ impl Document {
         let components = self.get_instructions("ftd#redirect");
 
         for v in &components {
+            let url = v
+                .get_interpreter_value_of_argument("url", &self.tdoc())
+                .and_then(|v| v.string(self.name.as_str(), 0).ok());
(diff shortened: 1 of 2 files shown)


ftd/src/interpreter/main.rs
```diff
@@ -1100,19 +1100,33 @@ impl Document {
             .collect_vec()
     }
 
-    pub fn get_redirect(&self) -> Option<(String, i32)> {
+    pub fn get_redirect(&self) -> Result<Option<(String, i32)>, super::Error> {
```
Heulitig, line 1103: Use this Result type
```suggestion
    pub fn get_redirect(&self) -> fastn_core::Result<Option<(String, i32)> {
```

harshdoesdev: changed Result -> ftd::interpreter::Result

PR #1295, opened 2023-09-11
cookie support added with `ftd.cookie`

harshdoesdev:
A new internal `ftd.cookie` record and library function with the same name has been added to enable cookie support in fastn.

Example:
```ftd
-- ftd.cookie user: user-name
value: Harsh Singh
secure: true

-- ftd.cookie user-2: user-name
value: Rithik Seth
secure: true

-- string $user-name: *$user.value

-- ftd.text: $user-name

-- ftd.text: Save User Cookie
$on-click$: $save-user(a = $user)

-- ftd.text: Update User Cookie
$on-click$: $save-user(a = $user-2)

-- ftd.text: Get User
$on-click$: $get-user($a = $user-name)

-- void save-user(a):
ftd.cookie a:

ftd.cookie.set(a)

-- void get-user(a):
string $a:

__args__.a.set(ftd.cookie.get("user-name"))
```

### DIFF
--- a/fastn-js/js/ftd.js
+++ b/fastn-js/js/ftd.js
@@ -254,3 +254,66 @@ ftd.local_storage = {
         localStorage.removeItem(key);
     }
 }
+
+ftd.cookie = {
+    _get_key(key) {
+        if (key instanceof fastn.mutableClass) {
+            key = key.get();
+        }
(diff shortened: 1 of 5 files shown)


amitu:
@harshdoesdev before writing code write documentation for new features, and ensure you get buy in from everyone.

PR #1334, opened 2023-09-19
`text-shadow` property added

harshdoesdev:
added the `text-shadow` property in Text Commons for defining shadow for the text itself instead of the container.

Example:
```ftd
-- ftd.shadow s:
x-offset.px: 1
y-offset.px: 1
color: #000000
blur.px: 4
;; text-shadow does not accepts any value for spread
;; so it will safely ignore `spread` even if it is defined.

-- ftd.text: Hello World
align-self: center
text-shadow: $s
```


amitu:
Should text shadow be part of `ftd.type` property, so that once you apply a role, shadow also comes along.

PR #1359, opened 2023-10-03
Create single .output dir for build and cache

harshdoesdev:
Create single .output dir for build and cache. This way they can be packed into a single artifact and only one http request will be required to upload/retrieve it.


harshdoesdev:
Better way of doing this, we can make the vercel-cache-helper create a single folder with the contents of build and cache folder and let it handle the placement of these folders in their respective locations during the decompression phase.

PR #1382, opened 2023-10-15
allow unverified certificate

amitu:
When connecting to PostgreSQL we currently support 1. unencrypted connection, and 2. verified certificate. 

Heroku does not allow unencrypted connection, and the certificate they issue is self signed (I think), so can not be verified. 

and to support Heroku PostgreSQL we have added `FASTN_PG_ALLOW_UNVERIFIED_CERTIFICATE`. If this environment variable is present, set to any value, unverified certificate would be be accepted. 

Using this environment variable is a bad idea and right certificate should be used in production using `FASTN_PG_CERTIFICATE` environment variable.

### DIFF
--- a/fastn-core/src/library2022/processor/pg.rs
+++ b/fastn-core/src/library2022/processor/pg.rs
@@ -5,30 +5,33 @@ async fn create_pool() -> Result<deadpool_postgres::Pool, deadpool_postgres::Cre
         recycling_method: deadpool_postgres::RecyclingMethod::Verified,
     });
     let runtime = Some(deadpool_postgres::Runtime::Tokio1);
-    match std::env::var("FASTN_PG_CERTIFICATE") {
-        Ok(cert) => {
-            let cert = tokio::fs::read(cert).await.unwrap();
-            let cert = native_tls::Certificate::from_pem(&cert).unwrap();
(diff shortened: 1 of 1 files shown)


amitu:
Having said that, I have not been able to figure out how to use the root certificate Heroku published.  As per their docs the certificates are available at [`/etc/ssl/certs/ca-certificates.crt`](https://devcenter.heroku.com/articles/heroku-postgres-enhanced-certificates#connect-to-a-heroku-postgres-database-with-enhanced-certificates). But if I download it and try to use it:

```sh
heroku ps:copy /etc/ssl/certs/ca-certificates.crt -a fastn-todo-app
export FASTN_PG_CERTIFICATE=ca-certificates.crt
fastn serve --edition=2023
```

I get the following error:

```txt
WARN: FASTN_PG_DANGER_ALLOW_UNVERIFIED_CERTIFICATE is set to true, this is not recommended for production use
thread 'actix-server worker 2' panicked at /Users/amitu/Projects/fastn/fastn-core/src/library2022/processor/pg.rs:62:61:
called `Result::unwrap()` on an `Err` value: Error { code: -50, message: "One or more parameters passed to a function were not valid." }
```

PR #1413, opened 2023-10-23
check for updates

siddhantk232:
see #1410 for more details.


fastn/src/main.rs
```diff
@@ -26,11 +26,25 @@ pub enum Error {
...
-    Ok(())
+
+    futures::try_join!(
+        fastn_core_commands(&matches),
+        check_for_update_cmd(&matches)
+    )?;
+
+    match std::env::var("FASTN_CHECK_FOR_UPDATES") {
+        Ok(val) => {
+            if val == "1" && !matches.get_flag("check-for-updates") {
```
amitu, line 41: If env is set, let's assume any value other than `false` is true.

fastn/src/main.rs
```diff
@@ -244,9 +258,46 @@ async fn fastn_core_commands(matches: &clap::ArgMatches) -> fastn_core::Result<(
...
+        .send()
+        .await?
+        .json()
+        .await?;
+
+    let current_version = version();
+
+    if release.tag_name != current_version {
+        println!(
+                "You are using fastn {}, and latest release is {}, visit https://fastn.com/install to learn how to upgrade.",
```
amitu, line 289: The url should be `https://fastn.com/install/`.

amitu:
Let’s also add a sub command, so people can use it to see if we are up to date. `fastn check-for-updates`

siddhantk232:
> I changed my mind. `fastn -c` (without any sub command) should check for updates and report. Should not show help text, which we show is there are no sub commands specified.

I have made changes to do this. Below is the detailed behaviour:

- In the case when `FASTN_CHECK_FOR_UPDATES` is set, it'll silently check for updates and will only log to the console if a new version is available.
- In the case when `-c` is used with a subcommand, it'll also silently check for updates and will only log to the console if a new version is available.
- In the case when there's only `-c` with no subcommand, it'll report if the user is using the latest version or if they need to do an update.

PR #1423, opened 2023-10-26
SQL processor

harshdoesdev:
Based on the discussion in this pull request (PR) at https://github.com/fastn-community/todo-app/pull/1, a new `sql` processor has been added. This processor is designed to handle various SQL databases, such as PostgreSQL, MySQL, and SQLite, and provides a unified interface for these databases. Users will set an environment variable `FASTN_DB_URL`, and the`sql` processor will check this URL. It will automatically determine the database type from the URL and select the appropriate processor for query processing.


amitu:
@harshdoesdev follow https://pypi.org/project/dj-database-url/ to know how the URLs are supposed to look like. Looks like you are making your own stuff here.

fastn-core/src/library2022/processor/sqlite.rs
```diff
@@ -176,64 +200,86 @@ fn extract_named_parameters(
...
+                    line_number,
+                });
             }
         }
     }
 
+    // Handle the last param if there was no trailing comma or space
+    if let State::PushParam = state {
+        println!("Here");
+        if !param_name.is_empty() {
```
Arpita-Jaiswal:
```suggestion
    if state.eq(State::PushParam) && !param_name.is_empty() {
```

harshdoesdev: Error: "Method 'eq' is not found for this enum," but can we use 'matches!' instead?

harshdoesdev: Eg: `if matches!(state, State::PushParam) && !param_name.is_empty() {`

fastn-core/src/library2022/processor/sqlite.rs
```diff
@@ -190,6 +169,30 @@ fn resolve_variable_from_headers(
...
+    doc: &ftd::interpreter::TDoc,
+    headers: &ftd::ast::HeaderValues,
+    line_number: usize,
+) -> ftd::interpreter::Result<Box<dyn rusqlite::ToSql>> {
+    resolve_variable_from_headers(param_name, param_type, doc, headers, line_number)
+        .or_else(|_| resolve_variable_from_doc(param_name, doc, line_number))
+        .map(|v| Box::new(v) as Box<dyn rusqlite::ToSql>)
+}
+
+#[derive(Debug)]
```
Arpita-Jaiswal, line 184:
```suggestion
#[derive(Debug, PartialEq)]
```
Use `PartialEq`to use `.eq()`

fastn-core/src/library2022/processor/sqlite.rs
Arpita-Jaiswal: @harshdoesdev https://github.com/fastn-stack/fastn/pull/1423#discussion_r1377094773

fastn-core/src/library2022/processor/sql.rs
```diff
@@ -7,45 +7,31 @@ pub struct DatabaseConfig {
...
     kind: ftd::interpreter::Kind,
     doc: &ftd::interpreter::TDoc<'_>,
+    config: &fastn_core::Config,
 ) -> ftd::interpreter::Result<ftd::interpreter::Value> {
     let db_config = fastn_core::library2022::processor::sql::get_db_config()?;
     let db_type = db_config.db_type.as_str();
 
     match db_type {
-        "pg" => Ok(fastn_core::library2022::processor::pg::process(value, kind, doc).await?),
+        "postgres" => Ok(fastn_core::library2022::processor::pg::process(value, kind, doc).await?),
```
Arpita-Jaiswal, line 32: Why have we changed the processor name? Also if it's needed to be changed, are we not supporting backward incompatibility?

harshdoesdev: Actually it is not the processor name but the schema part of the connection uri eg: "postgres://". I have not removed any processor, so backward compatibility is there. people can still use those processors, they just have to use the env var `FASTN_DB_URL` instead of `FASTN_PG_URL` or `db` header for Sqlite. the `db` header has been removed as it was requested in the above mentioned PR.

Arpita-Jaiswal (approved):

PR #1425, opened 2023-10-26
.env support

### LINKED ISSUE #1424: `.env` support
- if there's a .env file in CWD, parse it and set env vars from it
- create-package should generate a .gitignore file that contains .env file
- check if the .env file is checked in version control, panic or proceed only when
  FASTN_DANGER_ACCEPT_CHECKED_IN_ENV is set

siddhantk232:
closes #1424 

- if there's a .env file in CWD, parse it and set env vars from it
- create-package should generate a .gitignore file that contains .env
  file
- check if the .env file is checked in, panic and proceed only when
  FASTN_DANGER_ACCEPT_CHECKED_IN_ENV is set

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1458,6 +1458,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "dotenvy"
+version = "0.15.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


amitu:
In case `FASTN_DANGER_ACCEPT_CHECKED_IN_ENV` is on, issue a warning as well.

PR #1432, opened 2023-10-27
allowing accessing global variables inside ftd functions 

Heulitig:


Arpita-Jaiswal:
@Heulitig Is this done?

Heulitig:
> @Heulitig Is this done?

Not yet there are some cases which needs to be handled before merging the above PR which I found when testing on fastn.com.

amitu:
@Heulitig resolve conflict here. Is this ready otherwise?

PR #1439, opened 2023-10-31
fix: fastn -c

### LINKED ISSUE #1418: `fastn -c` failing when running from a non fastn package directory
```sh
fastn -c
FastnCoreError(HttpError(reqwest::Error { kind: Decode, source: Error("missing field `tag_name`", line: 1, column: 279) }))
```

siddhantk232:
fastn_core_commands should do nothing in case of no subcommand

closes #1418

### DIFF
--- a/fastn/src/main.rs
+++ b/fastn/src/main.rs
@@ -39,15 +39,7 @@ async fn async_main() -> Result<(), Error> {
         check_for_update_cmd(&matches)
     )?;
 
-    match std::env::var("FASTN_CHECK_FOR_UPDATES") {
-        Ok(val) => {
-            if val != "false" && !matches.get_flag("check-for-updates") {
-                check_for_update(false).await?;
-            }
-            Ok(())
-        }
-        Err(_) => Ok(()),
-    }
+    Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


fastn/src/main.rs
```diff
@@ -61,6 +53,10 @@ async fn fastn_core_commands(matches: &clap::ArgMatches) -> fastn_core::Result<(
     use colored::Colorize;
     use fastn_core::utils::ValueOf;
 
+    if matches.subcommand_name().is_none() {
```
siddhantk232, line 56: This change was required as call to `fastn_core_commands()` calls `fastn_core::Config::read()` which errors when used outside of package dir even when no subcommand used in the cli.

siddhantk232, line 56: This change was required as call to `fastn_core::Config::read()` made on [line 84](https://github.com/fastn-stack/fastn/pull/1439/files#diff-0d3e40f8767e766fe81a191eb5706ed7afa961d1ad16fc4bc47d103c5276d823L87) errors when used outside of a package dir even when this is not required when no subcommand is passed in the cli.

amitu (approved):

PR #1460, opened 2023-11-07
backwards compatibility for package-query

harshdoesdev:
## What's Changed

Based on [this](https://github.com/orgs/fastn-stack/discussions/1455) discussion, the `db` header has been reintroduced to maintain backward compatibility with the `package-query` processor.

## How It Works

When the `package-query` processor is called, it will attempt to read the database connection string from the `db` header. It will also show a warning message saying that the `package-query` processor has been deprecated. The `sql` processor remains unchanged, and something similar to the `db` header will be introduced once the configuration is finalized.


amitu:
@harshdoesdev can you deprecate the entire package-query processor as well?

fastn-core/src/library2022/processor/sqlite.rs
```diff
@@ -33,6 +33,44 @@ pub async fn process(
...
+    let use_db = match headers.get_optional_string_by_key("use", doc.name, value.line_number()) {
+        Ok(Some(k)) => Some(k),
+        _ => match headers
+            .get_optional_string_by_key("db", doc.name, value.line_number())
+            .ok()
+        {
+            Some(k) => {
+                println!("Warning: `db` header is deprecated, use `use` instead.");
+                k
+            }
```
Arpita-Jaiswal, line 45: Lets just use `db` header for this PR, we are not so finalised in the use of `use`. We can make this change, if finalised, later.

fastn-core/src/library2022/processor/sqlite.rs
```diff
@@ -33,6 +33,44 @@ pub async fn process(
...
+            None => None,
+        },
+    };
+
+    let sqlite_database_path = match use_db {
+        Some(k) => {
+            let mut db_path = camino::Utf8PathBuf::new().join(k.as_str());
```
Arpita-Jaiswal, line 52: Can we do this instead? I think this should work.

```suggestion
            let mut db_path = config.root.join(k.as_str());
```

amitu: This is better. All file access must happen with root as soon we are going to have multiple fastn packages in different root directories.

fastn-core/src/library2022/processor/package_query.rs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            Some(k) => {
+                fastn_core::library2022::utils::log_deprecation_warning(
+                    "`db` header is deprecated, use `use` instead.",
+                );
+                k
+            }
+            None => None,
+        },
+    };
```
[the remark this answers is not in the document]
harshdoesdev: Got it

fastn-core/src/library2022/processor/pg.rs
```diff
@@ -87,9 +87,16 @@ pub async fn process(
...
     doc: &ftd::interpreter::TDoc<'_>,
+    deprecated_usage: bool,
 ) -> ftd::interpreter::Result<ftd::interpreter::Value> {
     let (headers, query) = super::sqlite::get_p1_data("pg", &value, doc.name)?;
 
+    if deprecated_usage {
+        fastn_core::library2022::utils::log_deprecation_warning(
+            "`pg` has been deprecated, use `sql` processor instead.",
+        );
+    }
```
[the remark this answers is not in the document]
harshdoesdev: Ok

Arpita-Jaiswal (approved):

Arpita-Jaiswal (approved):

PR #1463, opened 2023-11-07
server error should show a message

amitu:
@Arpita-Jaiswal I have included the middleware code from https://github.com/robjtede/actix-web-lab and changed one line, instead of empty string, we show "500 server error".

I did this because during development I was seeing empty page on server errors, and I had to check the terminal to see if was a server error or a blank page.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -188,54 +188,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


amitu:
Let's merge this when test passes.

Arpita-Jaiswal (approved):

PR #1467, opened 2023-11-08
Prevent internal class names from being removed

harshdoesdev:
## What's the Issue?

The issue is that when using the `classes` header to set custom CSS class names, the class names of the kernel elements are being removed.

## How to Replicate

fastn code:

```ftd
-- ftd.column:
css: styles.css
classes: red

-- ftd.text: Hello World

-- end: ftd.column
```

Expected Output:

```html
<div class="ft_column red"><div>Hello World</div></div>
```

Actual Output:

```html
<div class="red"><div>Hello World</div></div>
```

## Proposed Solution

In `fastn-js/js/dom.js`, we can create a map of class names called `fastn_dom.InternalClass` that will be used by kernel components. Then, in the `fastn_utils.removeNonFastnClasses` function, we can check if the className that is going to be removed is not one of those class names.

### DIFF
--- a/fastn-js/js/dom.js
+++ b/fastn-js/js/dom.js
@@ -2,6 +2,12 @@ let fastn_dom = {};
 
 fastn_dom.styleClasses = "";
 
+fastn_dom.InternalClass = {
+    FT_COLUMN: 'ft_column',
+    FT_ROW: 'ft_row',
+    FULL: "full",
+};
+
 fastn_dom.codeData = {
     availableThemes: {},
     addedCssFile: []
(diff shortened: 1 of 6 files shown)


fastn-js/js/dom.js
```diff
@@ -2,6 +2,12 @@ let fastn_dom = {};
 
 fastn_dom.styleClasses = "";
 
+fastn_dom.InternalClass = {
+    FT_COLUMN: 'ft_column',
+    FT_ROW: 'ft_row',
+    FULL: "full",
```
Arpita-Jaiswal, line 8: Do you think `full` is a good name? Can you think some better name then this? @harshdoesdev @Heulitig

harshdoesdev: We think `ft_full_size` is a better name

Arpita-Jaiswal: Go ahead with this @harshdoesdev @Heulitig

Arpita-Jaiswal (approved):

PR #1472, opened 2023-11-09
Google Sheets Query support in `sql` processor

harshdoesdev:
## Background

We are adding support for Google Sheets [`Query`](https://support.google.com/docs/answer/3093343?hl=en) to the `sql` processor. You can find more information on this and participate in the related [`discussion`](https://github.com/orgs/fastn-stack/discussions/1450).

## How it Works

The [Google Charts API](https://developers.google.com/chart/interactive/docs/querylanguage) allows you to query a Google Sheet much like you would with an SQL database, though it currently only supports data retrieval queries such as "SELECT." Our primary objective is to enable users to provide the URL of their Google Sheet as the `db` parameter for the `sql` processor. Currently, only public sheets are supported, but we plan to add support for private sheets in the future.

Example:
```ftd

-- record user:
string Name:
integer Age:
string Gender:

-- user list users:
$processor$: pr.sql
db: https://docs.google.com/spreadsheets/d/1FOSZWtfQni0sNLLzbyhHGqzpZcRJx0_TLO8UGtnDXiY
sheet: Sheet1

SELECT *

-- ftd.text: $u.Name
for: $u in $users
```

## Todos
- [x] Basic Processor Setup
- [x] Google Charts API Integration 
- [x] Data type conversion
- [x] Getting a basic demo running
- [x] Resolve Named parameters
- [x] Move from CSV to JSON
- [x] Handle Formatted values
- [x] Query specific sheet by passing sheet name as `sheet`
- [ ] Documentation

## Updates
- For now I am going ahead with `sheet` and `db` instead of `$sheet$` and `$db$, we will deprecate these in a follow up PR and replace these with `$sheet$` and `$db$`.
- Skipping the `fbt` tests for now, will do manual testing instead. will add fbt tests in a follow up PR


fastn-core/src/library2022/processor/google_sheets.rs
```diff
@@ -0,0 +1,98 @@
...
+    if let Some(captures) = GOOGLE_SHEETS_ID_REGEX.captures(url) {
+        if let Some(id) = captures.get(1) {
+            return Some(id.as_str().to_string());
+        }
+    }
+
+    None
+}
+
+pub fn generate_google_sheet_url(google_sheet_id: &str) -> String {
```
Arpita-Jaiswal, line 16:
```suggestion
pub(crate) fn generate_google_sheet_url(google_sheet_id: &str) -> String {
```

fastn-core/src/library2022/processor/google_sheets.rs
```diff
@@ -0,0 +1,98 @@
+const GOOGLE_SHEET_API_BASE_URL: &str = "https://docs.google.com/a/google.com/spreadsheets/d";
+
+static GOOGLE_SHEETS_ID_REGEX: once_cell::sync::Lazy<regex::Regex> =
+    once_cell::sync::Lazy::new(|| regex::Regex::new(r"/spreadsheets/d/([a-zA-Z0-9-_]+)").unwrap());
+
+pub fn extract_google_sheets_id(url: &str) -> Option<String> {
```
Arpita-Jaiswal, line 6:
```suggestion
pub(crate) fn extract_google_sheets_id(url: &str) -> Option<String> {
```

Arpita-Jaiswal: Convert the functions that are not going to used outside crate as `pub(crate)`

harshdoesdev: Got it, thanks

fastn-core/src/library2022/processor/google_sheets.rs
```diff
@@ -0,0 +1,298 @@
...
+        if let Some(stripped) = value.strip_prefix('$') {
+            return resolve_variable_from_doc(stripped, doc, line_number);
+        }
+    }
+
+    let param_value: String = match (param_type, &header.value) {
+        ("STRING", ftd::ast::VariableValue::String { value, .. }) => escape_string_value(value),
+        ("INTEGER", ftd::ast::VariableValue::String { value, .. }) => value.to_string(),
+        ("DECIMAL", ftd::ast::VariableValue::String { value, .. }) => value.to_string(),
+        ("BOOLEAN", ftd::ast::VariableValue::String { value, .. }) => value.to_string(),
```
Arpita-Jaiswal, line 117:
```suggestion
        ("INTEGER", ftd::ast::VariableValue::String { value, .. }) |
        ("DECIMAL", ftd::ast::VariableValue::String { value, .. }) |
        ("BOOLEAN", ftd::ast::VariableValue::String { value, .. }) => value.to_string(),
```

fastn-core/src/library2022/processor/google_sheets.rs
```diff
@@ -26,32 +43,332 @@ pub(crate) fn prepare_query_url(url: &str, query: &str) -> String {
...
+#[derive(Debug, serde::Deserialize)]
+pub(crate) struct DataValue {
+    v: serde_json::Value,
+    #[serde(default)]
+    f: Option<String>,
+}
+
+#[derive(Debug, serde::Deserialize)]
+pub(crate) struct DataTable {
+    cols: Vec<DataColumn>,
```
amitu, line 68: Rename this to `header`.

amitu: Or `schema`.

amitu: @harshdoesdev rename `cols` to `schema`.

fastn-core/src/library2022/processor/google_sheets.rs
```diff
@@ -78,7 +78,7 @@ pub(crate) struct QueryResponse {
     // req_id: String,
     // status: String,
     // sig: String,
-    table: DataTable,
+    schema: DataSchema,
```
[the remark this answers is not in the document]
harshdoesdev: Got it

fastn-core/src/library2022/processor/sql.rs
```diff
@@ -4,19 +4,25 @@ pub struct DatabaseConfig {
...
 }
 
+impl DatabaseConfig {
+    pub fn new(db_url: String, db_type: String) -> DatabaseConfig {
+        DatabaseConfig { db_url, db_type }
+    }
+}
+
 pub(crate) fn get_db_config() -> ftd::interpreter::Result<DatabaseConfig> {
     let db_url = std::env::var("FASTN_DB_URL").expect("FASTN_DB_URL is not set");
```
amitu, line 14: Let's convert this to Result as well and not panic if its not too difficult.

fastn-core/src/library2022/processor/google_sheets.rs
```diff
@@ -0,0 +1,612 @@
...
+                _ => {
+                    return Err(ftd::interpreter::Error::ParseError {
+                        message: format!("Can't parse to boolean, found: {}", &data_value.v),
+                        doc_id: doc.name.to_string(),
+                        line_number,
+                    })
+                }
+            },
+        },
+        _ => unimplemented!(),
```
amitu, line 318: No panic allowed, return Result.

fastn-core/src/library2022/processor/google_sheets.rs
```diff
@@ -0,0 +1,612 @@
+const GOOGLE_SHEET_API_BASE_URL: &str = "https://docs.google.com/a/google.com/spreadsheets/d";
```
amitu, line 1: This file has more than just the processor stuff, let's move it to `fastn-core/src/google_sheets.rs`. It will clean up some full path references, they are really long.

amitu:
@Arpita-Jaiswal can you take a look at the fastn-core/src/library2022/processor/google_sheets.rs, there is a lot of stuff happening there and I want your opinion.

fastn-core/src/library2022/processor/google_sheets.rs
harshdoesdev: Got it

harshdoesdev:
@amitu I have temporarily removed the timestamp conversion for now. I have added a new optional processor param named `sheet` where the user can specify the name of the sheet they want to query.

Example:
```ftd
-- user list users:
$processor$: pr.sql
db: https://docs.google.com/spreadsheets/d/1FOSZWtfQni0sNLLzbyhHGqzpZcRJx0_TLO8UGtnDXiY
sheet: Users
name: John Doe

SELECT * WHERE A = $name::STRING
```

amitu:
The sheet argument conflicts if the sheet is a column name. Should we name processor arguments as `$<arg-name>$`, eg `$sheet$: <sheet name>` and same with `db`.

harshdoesdev:
> The sheet argument conflicts if the sheet is a column name. Should we name processor arguments as `$<arg-name>$`, eg `$sheet$: <sheet name>` and same with `db`.

I agree, but it should not conflict with the column name because these words are rarely used, and this looks cleaner without '$'. However, this is just my opinion.

amitu:
It looks cleaner, but if conflict happens user will have no choice. `sheet` and `db` are both common enough terms. Let's switch to `$foo$` for all processor arguments as this is a general problem and we need a general / consistent solution.

harshdoesdev:
> It looks cleaner, but if conflict happens user will have no choice. `sheet` and `db` are both common enough terms. Let's switch to `$foo$` for all processor arguments as this is a general problem and we need a general / consistent solution.

Yes, although these argument names need not be the same as the column name, many users choose to name them to match the column name. Therefore, I agree that it can lead to conflicts.

Heulitig (approved):

PR #1481, opened 2023-11-11
cargo publish during release

amitu:
In this PR I am trying to enable cargo publish command on each release so people can install `fastn` using `cargo install fastn` as well. 

We probably do not recommend that (as it recompiles everything, and is slower), but it would be good to have our package history on crates.io, and some people are more trusting of `cargo install` way than running shell scripts to install things.

References: 
- [CARGO_REGISTRY_TOKEN docs](https://doc.rust-lang.org/cargo/reference/config.html?highlight=CARGO_REGISTRY_TOKEN#credentials)
- [Using secrets in GitHub Actions](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -148,3 +148,29 @@ jobs:
           tag: ${{ github.event.inputs.releaseTag }}
           prerelease: ${{ github.event.inputs.productionRelease && github.event.inputs.productionRelease == 'false' }}
           bodyFile: .github/RELEASE_TEMPLATE.md
+  publish-crate:
+    name: Publish our crate on crates.io
+    runs-on: ubuntu-latest
+    needs: [create-release]
+    steps:
+      - uses: actions/checkout@main
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+            ftd/target
+            fifthtry_content/target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+      - name: Install stable toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+      - name: Publish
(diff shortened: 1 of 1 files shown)


amitu:
I have so far created a token and published it. I am going to publish 0.4 manually using `cargo publish` on my machine (by deleting login token, and using the environment variable and the token I have uploaded on Github secrets) to make sure 0.4 release happens cleanly. Once that is done we can merge this PR and test it.

.github/workflows/release.yml
```diff
@@ -148,3 +148,31 @@ jobs:
...
+          profile: minimal
+          toolchain: stable
+          override: true
+      - name: Publish
+        uses: gmiam/rust-musl-action@9e6a37bf27ecfffb6b92240ea276bea5487fa15d
+        continue-on-error: false
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+        with:
+          args: cargo publish
```
amitu, line 178:
```suggestion
          args: cargo publish -p fastn
```

amitu:
When trying to publish from local I am getting this error:

```sh
cargo publish -p fastn
    Updating crates.io index
error: all dependencies must have a version specified when publishing.
dependency `fastn-cloud` does not specify a version
Note: The published dependency will use the version from crates.io,
the `path` specification will be removed from the dependency declaration.
```

amitu:
[How to perform nextest releases](https://github.com/fastn-stack/fastn/pull/1481#how-to-perform-nextest-releases)

amitu:
[create-gh-release-action](https://github.com/taiki-e/create-gh-release-action)

amitu:
[cargo-dist book](https://axodotdev.github.io/cargo-dist/book/)

amitu:
[Fully Automated Releases for Rust Projects](https://blog.orhun.dev/automated-rust-releases/)

PR #1501, opened 2023-11-14
Fix SO error in debug builds

### LINKED ISSUE #1067: fastn binary crashes if built in debug mode
Release build works, but debug build fails with stack overflow error. Because of this development is affected, building in release mode takes longer than debug mode, so edit-compile-test cycle is slower. Also we should fix this just because it is a smell, someday release will also start crashing, or may be it is already crashing on some operating systems etc.

siddhantk232:
closes #1067


amitu (approved):
You my friend rock!

PR #1512, opened 2023-11-16
Wrapped internal fastn.js classes in an IIFE to avoid namespace conflicts

harshdoesdev:
Wrapped internal fastn.js classes in an IIFE to avoid namespace conflicts.


harshdoesdev:
I have tested it on fastn.com (checked search and notification), and it is working.

Arpita-Jaiswal (approved):

PR #1518, opened 2023-11-20
Rename `is-login` -> `is-logged-in`

siddhantk232:
This PR addresses some suggestions made in #1485 after merge.


fastn-core/src/library2022/processor/user_details.rs
```diff
@@ -1,46 +1,42 @@
+use fastn_core::auth;
+
```
amitu, line 2: We do not do this. Use full paths.

amitu (approved):

Arpita-Jaiswal (approved):

PR #1519, opened 2023-11-20
feat: `user_id` cookie for authenticated user

siddhantk232:
Set a `user_id` cookie when user is authenticated. This cookie stores an encrypted json string, eg:

```js
{
  'login': 'siddhantk232', 
  'id': 36407043, 
  'name': 'Siddhant Kumar'
}
```

All cookies are forwarded to the proxy backend configured using the `endpoint` package header so we don't need `user-id: true` option now.

### DIFF
--- a/fastn-core/src/commands/serve.rs
+++ b/fastn-core/src/commands/serve.rs
@@ -332,6 +332,7 @@ pub async fn serve(
 
             // TODO: read app config and send them to service as header
             // Adjust x-fastn header from based on the platform and the requested field
+            // this is for fastn.app
             if let Some(user_id) = conf.get("user-id") {
                 match user_id.split_once('-') {
                     Some((platform, requested_field)) => {
@@ -346,6 +347,32 @@ pub async fn serve(
                         }
                     }
                     _ => return Ok(fastn_core::unauthorised!("invalid user-id provided")),
+                }
+            }
+
+            if config.package.user_id {
+                let provider = "github";
+                if let Some(token) =
+                    fastn_core::auth::get_user_data_from_cookies(provider, "token", &cookies)
+                        .await
+                        .ok()
+                        .flatten()
+                {
+                    match fastn_core::auth::github::apis::user_details(token.as_str()).await {
+                        Ok(user) => {
+                            // x-fastn-user-id is github_id
+                            conf.insert("X-FASTN-USER-ID".to_string(), format!("{}", user.id));
+                        }
+                        Err(_) => {
+                            tracing::error!(
+                                "Failed to pass used-id to endpoint. Invalid github access_token"
+                            );
+                        }
+                    }
+                } else {
+                    tracing::error!(
+                        "Failed to get token from {provider} cookie. User is not authenticated"
+                    );
                 }
             }
 
--- a/fastn-core/src/commands/serve.rs
+++ b/fastn-core/src/commands/serve.rs
@@ -350,32 +350,6 @@ pub async fn serve(
                 }
             }
 
-            if config.package.user_id {
-                let provider = "github";
-                if let Some(token) =
-                    fastn_core::auth::get_user_data_from_cookies(provider, "token", &cookies)
-                        .await
-                        .ok()
-                        .flatten()
-                {
-                    match fastn_core::auth::github::apis::user_details(token.as_str()).await {
-                        Ok(user) => {
-                            // x-fastn-user-id is github_id
-                            conf.insert("X-FASTN-USER-ID".to_string(), format!("{}", user.id));
-                        }
-                        Err(_) => {
-                            tracing::error!(
-                                "Failed to pass used-id to endpoint. Invalid github access_token"
-                            );
-                        }
-                    }
-                } else {
-                    tracing::error!(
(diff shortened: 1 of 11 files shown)


fastn-core/src/commands/serve.rs
```diff
@@ -349,6 +350,32 @@ pub async fn serve(
...
 
+            if config.package.user_id {
+                let provider = "github";
+                if let Some(token) =
+                    fastn_core::auth::get_user_data_from_cookies(provider, "token", &cookies)
+                        .await
+                        .ok()
+                        .flatten()
+                {
+                    match fastn_core::auth::github::apis::user_details(token.as_str()).await {
```
amitu, line 361: Making a http request for every request is expensive. Can you cache the user-id somewhere? A signed cookie is what we prefer.

amitu (changes requested):

fastn-core/src/commands/serve.rs
siddhantk232: This, I think would be more secure than just passing `x-fastn-user-id` header to the proxy backend.

siddhantk232: I'd like to learn more about why you'd suggest using http header over a cookie. In my opinion, using a cookie would be much more convenient as we'd be able to remove all this code that creates custom header (`x-fastn-user-id`) if the cookie is present. And, caching wouldn't be needed if we store `user-id` as a signed cookie.

amitu: My thinking it this: the header is better if the information is only passed from fastn to django, if it is a cookie then it should go till browser.

siddhantk232:
I have changed this implementation. Now we have a `user_id` encrypted cookie. [All cookies are forwarded to the proxy backend currently](https://github.com/fastn-stack/fastn/blob/main/fastn-core/src/proxy.rs#L97) so we don't need `user-id: true` package header for now.

I have already tested decrypting the cookie ([encrypted using the magic_crypt](https://github.com/fastn-stack/fastn/blob/main/fastn-core/src/auth/utils.rs#L11)) in python.

@amitu

amitu:
Some questions that are coming to my mind:

1. Should we rename the cookie to indicate this is Github Data? 
2. Do we need both `id` and `username`? Think from "what do we need on django side?", we only need one identifier, and I think `username` is a superior identifier. I can imagine the use of `id` as well, if you application really needs it? Basically I am thinking of reducing the cookie data size. 

@siddhantk232, @Arpita-Jaiswal what do you think?

siddhantk232:
> Some questions that are coming to my mind:
> 
> 1. Should we rename the cookie to indicate this is Github Data?
> 2. Do we need both `id` and `username`? Think from "what do we need on django side?", we only need one identifier, and I think `username` is a superior identifier. I can imagine the use of `id` as well, if you application really needs it? Basically I am thinking of reducing the cookie data size.
> 
> @siddhantk232, @Arpita-Jaiswal what do you think?

1. I agree. I'll rename this to `github_user` as it stores the user info (name and username) not just the id?
2. The `id` is also available in the `github` cookie (along with `access_token`) so I'llremove it from here. This will leave only username (`login`) and `name` in this cookie.

amitu:
Sounds good, let's do these two changes. Tell me more about the `github` cookie.

siddhantk232:
> Sounds good, let's do these two changes. Tell me more about the `github` cookie.

The old impl uses the `github` cookie which contains the `username` and github `access_token`. This is used in some places so I haven't touched it.
The `user_details` processor also uses this `github` cookie and uses the `access_token` to query user details from the github api. But, we can replace this cookie with our new `github_user` cookie which will also remove the need for caching for this processor.

In future, I expect the auth handler as well as all the helpers to get github related data (that require the access_token) to be in a separate auth package.

amitu:
If we had session we would have been storing all this on server side, in session store. Can we do a quick DB integration, create session table if it does not exist, and put all that there? Or else I am okay with putting all github related data in one signed cookie.

amitu:
Actually if we can use session table created by django from fastn it would be pretty cool, then django session stuff will just work.

siddhantk232:
> If we had session we would have been storing all this on server side, in session store. Can we do a quick DB integration, create session table if it does not exist, and put all that there? Or else I am okay with putting all github related data in one signed cookie.

I have changed the the code to use a single signed cookie (`github`) to store the information (github username, full name and access_token).

FWIW I did try to setup sessions in fastn yesterday using [`actix_session`](https://docs.rs/actix-session/latest/actix_session/) and it's not that easy. We'd have to extend [`fastn_core::http::Request`](https://github.com/fastn-stack/fastn/blob/94463f0af2e0ad4b4802c467df7c24b02e30b907/fastn-core/src/http.rs#L77) to be able to use it in our auth handlers.

I'd rather focus on getting the auth to work in slides.fifthtry.com using django middlewares for now. We will definitely add sessions in fastn too. Maybe we can write a custom [fastn auth backend for django](https://docs.djangoproject.com/en/4.2/topics/auth/customizing/#writing-an-authentication-backend) in future to integrate with django rather than using django specific models in fastn.

@amitu

PR #1542, opened 2023-11-30
preserve processor cookies

### LINKED ISSUE #1538: http processor loses cookies
When we make a request using [http processor](https://fastn.com/http/), if the response included any cookie, we do not propagate it to the browser. 

## Expected Behavior

Since we do send all cookies to upstream, it makes sense that upstream can also create cookies.

## Current Behavior

Upstream cookies are lost.

## Possible Solution

Store the cookies from each processor call in the Config or some such global structure, and pass it along the response.

Call a http processor that sets a cookie, see if cookie is found in browser. 

## Context

We are building slides project which is using django as backend. Django is setting a few cookies to keep track of things, and those cookies gets set if the request is directly forwarded to django via endpoint feature, but if django tries to set the cookie via http processors they are getting lost. 

## Your Environment

- Operating System: Mac
- Web Browser: Safari

amitu:
This PR fixes #1538.

### DIFF
--- a/fastn-core/src/commands/serve.rs
+++ b/fastn-core/src/commands/serve.rs
@@ -242,7 +242,7 @@ pub async fn serve(
 
     let path: camino::Utf8PathBuf = req.path().replacen('/', "", 1).parse()?;
 
-    Ok(if path.eq(&camino::Utf8PathBuf::new().join("FASTN.ftd")) {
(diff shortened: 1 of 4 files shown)


amitu:
@siddhantk232 I have written the code without testing it. Do take a look at it, and let me know what you think?

PR #1543, opened 2023-11-30
use user's primary email if public email is null

### LINKED ISSUE #1540: github cookie can contain null email
As reported in this thread: [Github user email is null, despite user:email scope](https://stackoverflow.com/questions/35373995/github-user-email-is-null-despite-useremail-scope), we are getting email as null in github cookie set by `fastn`. 

## Expected Behavior

As reported, the Github API does not return the email data in the user data API even if we have access to email, and in such cases we have to call the email api to get the email.

## Current Behavior

We blindly set email returned by user data api. 

## Possible Solution

Handle the email null case by calling another API.

Configure Github stuff, and then Login with Github, with an account that has made their email hidden in their Github Profile. Inspect the cookie content, email will be null.

## Context

We are using it for our slides project and our middleware handles missing email, but it's causing downstream issues. We want email address, and user has granted email address so we should not get null values.

## Your Environment

- Operating System: Mac
- Web Browser: Safari

siddhantk232:
Fixes #1540

### DIFF
--- a/fastn-core/src/auth/github/apis.rs
+++ b/fastn-core/src/auth/github/apis.rs
@@ -138,12 +138,27 @@ pub async fn is_user_sponsored(
     }
 }
 
-// TODO: It can be stored in the request cookies
 pub async fn user_details(access_token: &str) -> fastn_core::Result<GhUserDetails> {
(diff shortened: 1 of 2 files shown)


amitu:
We will merge this after #1541 goes through.

PR #1547, opened 2023-12-01
option to disable cache in http_get_with_cookie()

### LINKED ISSUE #1539: NOT_FOUND_CACHE should not be used by http.processor
To not keep re-looking a dependency that is missing we have implemented a in memory `NOT_FOUND_CACHE`, where if a URL returns 404 once, we store it and do not retry it.

This is done a http helper util level, which is also used in http processor and if upstream returns 404, that too gets cached. 

## Expected Behavior

Since the upstream is getting cookies etc, it may return 404 for some URL meant only for currently viewing user, caching it for all users is wrong. 

## Current Behavior

Everything get's cached, and we have to restart fastn server to purge the in-memory cache. Till that time everyone keeps getting 404 for that URL.

## Possible Solution

We split our utility to have different behaviour for downloading dependencies and for doing http processor. 

Configure a http processor to point to any URL that returns 404, fastn will only make one request to that URL. 

## Context

We are working on slides, and are using `get_or_404()` helper to return 404 error in case a row is not found. This workflow is broken as the row may be missing in one request and found in next (as it was created in between, or some permission related thing changed between the two requests).

## Your Environment

- Operating System: Mac
- Web Browser: Safari

siddhantk232:
- disable `NOT_FOUND_CACHE` cache for the `http` processor

Fixes #1539

### DIFF
--- a/fastn-core/src/http.rs
+++ b/fastn-core/src/http.rs
@@ -445,7 +445,7 @@ pub(crate) async fn http_post_with_cookie(
 }
 
 pub(crate) async fn http_get(url: &str) -> fastn_core::Result<Vec<u8>> {
-    http_get_with_cookie(url, None, &std::collections::HashMap::new())
+    http_get_with_cookie(url, None, &std::collections::HashMap::new(), true)
         .await?
         .0
 }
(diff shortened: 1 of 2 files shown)


fastn-core/src/http.rs
```diff
@@ -513,7 +514,11 @@ pub(crate) async fn http_get_with_cookie(
             res.text().await
         );
         tracing::error!(url = url, msg = message);
-        NOT_FOUND_CACHE.write().insert(url.to_string());
+
+        if use_cache {
+            NOT_FOUND_CACHE.write().insert(url.to_string());
+        }
```
Arpita-Jaiswal, line 520:
```suggestion
       NOT_FOUND_CACHE.write().insert(url.to_string());
```
We have already checked this condition above.

siddhantk232: shouldn't we not insert entries from http processors if the `use_cache` is `false`? This will reduce memory usage

Arpita-Jaiswal: I have misread the code. I thought it was inside `if use_cache && NOT_FOUND_CACHE.read().contains(url) {` block. You are right.

Arpita-Jaiswal (approved):

PR #1603, opened 2023-12-13
`fastn test` command

Arpita-Jaiswal:
https://github.com/orgs/fastn-stack/discussions/1586


fastn-core/src/config/mod.rs
```diff
@@ -716,10 +716,30 @@ impl Config {
         Ok(documents)
     }
 
+    pub(crate) async fn get_test_files(
+        &self,
+        package: &fastn_core::Package,
+    ) -> fastn_core::Result<Vec<fastn_core::File>> {
```
amitu, line 722: Ideally this function should itself be in `fastn_core::commands::test` module.

amitu (approved):

PR #1612, opened 2023-12-16
Multi Auth

siddhantk232:
This PR adds emailpassword login (sign in using email and password) on top of the current github oauth. Both github oauth and emailpassword strategy create a `fastn_user` (database table). See [`multiauth.sql`](https://github.com/fastn-stack/fastn/pull/1612/files#diff-55745e5392b5532faf69705ec92fcefb857c66499f76fc85919d0322d8b3b0b6) for more details on the required tables for authentication.

- you're not allowed to login simulatenously with github and emailpassword.
  It'll be possible in future to merge two accounts by logging in to two
  accounts.
- You can either use emailpassword or github. The password set in emailpassword
  step will not be considered when logged in using github
- tested with login, logout, github login/logout, and the `user-details`
  processors. The `user-details` processor requires no change in existing
  codebases. Expect bugs in other places.

### TODO

- DONE ~Refactor code that makes database queries in `emailpassword.rs`. The [`mod db`](https://github.com/fastn-stack/fastn/pull/1612/files#diff-132cc066fdf02e7f24c5b7045d03df1a554b6df1c0428d92fc08cf2fe35c75d1R280) uses code copied from the `sql` processor. Ideally they should share the db pool. This will be fixed in a new PR~.
- Write tests. It is very hard without doing major refactors so it is ignored in this PR so as to quickly move to figuring out sending emails. Tests will also be added for this code in a new PR later.


multiauth.sql
```diff
@@ -0,0 +1,30 @@
+-- TODO: move to some place else
+
+CREATE TABLE IF NOT EXISTS fastn_user (
+    id UUID PRIMARY KEY,
```
amitu, line 4: Let's make this `integer` instead.

multiauth.sql
```diff
@@ -0,0 +1,30 @@
...
+    email VARCHAR(254) NOT NULL,
```
amitu, line 5: This should also be TEXT. Or we create a [new domain](https://stackoverflow.com/questions/62113787/what-should-be-the-data-type-of-field-email-in-postgresql-database-in-pgadmin).

multiauth.sql
```diff
@@ -0,0 +1,30 @@
...
+    username TEXT,
+    password TEXT,
+    name TEXT NOT NULL,
+    created_on TIMESTAMP
```
amitu, line 9: This should be of `timestamp with time zone` type.

multiauth.sql
```diff
@@ -0,0 +1,30 @@
...
+);
+
+CREATE TABLE IF NOT EXISTS fastn_session (
+    id UUID PRIMARY KEY,
```
amitu, line 13: All `id`s should be `integer`.

multiauth.sql
```diff
@@ -0,0 +1,30 @@
...
+    user_id UUID REFERENCES fastn_user(id),
+    created_on TIMESTAMP
+);
+
+CREATE TABLE IF NOT EXISTS fastn_oauthtoken (
+    id UUID PRIMARY KEY,
+    session_id UUID NOT NULL REFERENCES fastn_session(id) ON DELETE CASCADE,
+    token text NOT NULL,
+    provider text NOT NULL,
+    created_on TIMESTAMP DEFAULT now()
```
amitu, line 23: This should also contain `updated_on`.

Do we also have to store refresh token? And the token validity period?

multiauth.sql
amitu: Actually study these tables, ours should look similar to them: https://github.com/pennersr/django-allauth/blob/main/allauth/account/models.py

fastn-core/src/auth/emailpassword.rs
```diff
@@ -0,0 +1,371 @@
+use argon2::{Argon2, PasswordHasher, PasswordVerifier};
```
amitu, line 1: Let's rename the module to `email_password`.

fastn-core/src/auth/emailpassword.rs
```diff
@@ -0,0 +1,371 @@
...
+}
+
+impl Token {
+    fn from_row(row: &tokio_postgres::Row) -> Self {
+        Token {
+            session_id: row.get("session_id"),
+            token: row.get("token"),
+            provider: row.get("provider"),
+        }
+    }
```
amitu, line 17: I prefer diesel so we can get compile time checking for these types.

fastn-core/src/auth/emailpassword.rs
```diff
@@ -0,0 +1,371 @@
...
+            &[&user.username, &user.email],
+        )
+        .await
+        .unwrap();
+
+    let user_count: i64 = res[0].get(0);
+
+    if user_count > 0 {
+        // TODO: use correct HTTP status code
+        return fastn_core::http::api_error("user already exists");
```
amitu, line 54: Error returned by these APIs must match ones returned by `fastn.django` form errors.

fastn-core/src/auth/emailpassword.rs
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    // encrypt password
```
amitu, line 57: You are hashing password, not encrypting it.

fastn-core/src/auth/emailpassword.rs
```diff
@@ -0,0 +1,371 @@
...
+    let salt =
+        argon2::password_hash::SaltString::generate(&mut argon2::password_hash::rand_core::OsRng);
+
+    let argon2 = argon2::Argon2::default();
+
+    let hashed_password = argon2
+        .hash_password(user.password.as_bytes(), &salt)
+        .unwrap()
```
amitu, line 65: Let's not do unwrap()

fastn-core/src/auth/emailpassword.rs
```diff
@@ -0,0 +1,371 @@
...
+        user.email.as_str(),
+        user.username.as_str(),
+        user.name.as_str(),
+        hashed_password.as_str(),
+    )
+    .await?;
+
+    tracing::info!("fastn_user created. user: {:?}", &user);
+
+    fastn_core::http::api_ok(user)
```
amitu, line 84: On success we return redirect to the right place.

fastn-core/src/auth/emailpassword.rs
```diff
@@ -0,0 +1,371 @@
...
+
+    let client = db::get_client().await?;
+
+    let user = client
+        .query(
+            "select id, username, email, name, password from fastn_user where email = $1 limit 1",
+            &[&body.email],
+        )
+        .await
+        .unwrap_or(vec![]);
```
amitu, line 114: Why unwrap?

fastn-core/src/auth/emailpassword.rs
```diff
@@ -0,0 +1,371 @@
...
+
+    if user.is_empty() {
+        return fastn_core::http::api_error("invalid payload");
```
amitu, line 117: Error handling has to change everywhere.

fastn-core/src/auth/emailpassword.rs
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    let user = user.first().expect("vec of length must have first");
```
amitu, line 120: You are returning 500 if email does not exist. For every 500 you will have to wake up and fix the problem.

fastn-core/src/auth/emailpassword.rs
```diff
@@ -0,0 +1,371 @@
...
+
+    if password_match.is_err() {
+        return fastn_core::http::api_error("invalid payload");
+    }
+
+    let session_id = uuid::Uuid::new_v4();
+
+    let affected = create_session(&session_id, &user_id).await.unwrap_or(0);
+
+    tracing::info!("session created. Rows affected: {}", &affected);
```
amitu, line 137: Let's do more informative login, for example log the email address, actually user id as email is PII and we do not want to log PIIs unless necessary.

fastn-core/src/auth/emailpassword.rs
```diff
@@ -0,0 +1,371 @@
...
+
+/// insert user if it not exists in the database returning the rows affected
+/// email and username is unique in fastn_user table
+pub async fn upsert_user(
+    user_id: &uuid::Uuid,
+    email: &str,
+    username: &str,
+    name: &str,
+    hashed_password: &str,
+) -> fastn_core::Result<fastn_core::auth::FastnUser> {
```
amitu, line 152: I am not fond of this kind of functions. At every call site you know precisely if the user must exist and it must be an update user call, or user must not exist and it is insert call. You want things to fail as soon as possible, if code has a bug it must crash the server tonight, instead of silently corrupting data for years.

fastn-core/src/auth/emailpassword.rs
```diff
@@ -0,0 +1,371 @@
...
+
+    tracing::info!("fastn_user created. user: {:?}", &user);
+
+    fastn_core::http::api_ok(user)
+}
+
+pub(crate) async fn login(
+    req: &crate::http::Request,
+    next: String,
+) -> fastn_core::Result<fastn_core::http::Response> {
```
amitu, line 90: I would split all the HTTP handling, eg extracting data from `fastn_core::http::Request`[1], and creating `fastn_core::http::Response`, from the actual business logic, eg if tomorrow I have to call `login` from CLI either I will have to create fake http request object, extract information from http response, or refactor this code then. It's better to put them in separate places to begin with. 

[1]: do not use `crate`, if I am looking at this PR right now, I do not want to make it a puzzle finding task to figure out what `crate` you are referring to. Write code to make it easier for code reader, code gets read 100s of times, every single time the reader will have to do mental translation of `crate` to `fastn_core`, and if they ever make mistake it will cause confusion.

fastn-core/src/auth/emailpassword.rs
```diff
@@ -0,0 +1,371 @@
...
+
+    Ok(fastn_core::auth::FastnUser::from_row(
+        rows.first().expect("vec of length 1 must have first()"),
+    ))
+}
+
+// TODO: this is borrowed from pg.rs
+// pg.rs and this mod should use the same code
+pub mod db {
+    async fn create_pool() -> Result<deadpool_postgres::Pool, deadpool_postgres::CreatePoolError> {
```
amitu, line 281: This a fastn_core global functionality, a lot of code may need db access tomorrow, let's move this to `fastn_core::db`.

fastn-core/src/auth/mod.rs
```diff
@@ -2,17 +2,40 @@ pub(crate) mod config;
...
+use std::str::FromStr;
+
+pub const COOKIE_NAME: &str = "session";
+
+#[derive(Debug, serde::Deserialize, serde::Serialize)]
+pub struct FastnUser {
+    pub username: String,
+    pub id: uuid::Uuid,
+    pub name: Option<String>,
+    pub email: Option<String>,
```
amitu, line 18: Why is email optional?

fastn-core/src/auth/mod.rs
```diff
@@ -110,3 +139,27 @@ pub async fn get_auth_identities(
     };
     Ok(matched_identities)
 }
+
+async fn set_session_cookie_and_end_response(
+    req: &fastn_core::http::Request,
+    session_id: uuid::Uuid,
+    next: String,
+) -> fastn_core::Result<fastn_core::http::Response> {
```
amitu, line 147: We have added a test framework, check if this code works with it. This code should be fully end to end tested using our test framework.

fastn-core/src/auth/routes.rs
```diff
@@ -11,19 +13,32 @@ pub async fn login(
...
-pub fn logout(
+pub async fn logout(
     req: &fastn_core::http::Request,
     next: String,
 ) -> fastn_core::Result<fastn_core::http::Response> {
     // TODO: Refactor, Not happy with this code, too much of repetition of similar code
+
+    if let Some(session_id) = req.cookie(fastn_core::auth::COOKIE_NAME) {
+        let session_id =
+            uuid::Uuid::from_str(session_id.as_str()).expect("cookie contains valid uuid");
```
amitu, line 31: We can not crash request (and wake up the entire developer team if some random user sends a request with malformed cookie). Just internalise every unwrap/expect means some developer have to wake up at 3AM and check the production log about why the site is giving server errors.

amitu (changes requested):

siddhantk232:
Thanks @amitu for the review. I'll fix the issues raised, add tests and, integrate [diesel orm](https://github.com/diesel-rs/diesel) as you suggested.

multiauth.sql
amitu: I just saw this one: https://github.com/loco-rs/loco/blob/master/starters/saas/migration/src/m20220101_000001_users.rs. I kind of prefer allauth approach tho.

siddhantk232: I am liking allauth too. They have SocialAccount and other models that we can take inspiration from (for oauth providers)

.envrc
```diff
@@ -0,0 +1,2 @@
+export DATABASE_URL=postgresql://fastn:fastn@172.17.0.1:5432/fastn
+use flake
```
amitu, line 2: This is causing problem on my machine, I also have nix setup so it starts, but I do not use nix for developer setup. Can you put it in `.gitignore`?

diesel.toml
```diff
@@ -0,0 +1,9 @@
+# For documentation on how to configure this file,
+# see https://diesel.rs/guides/configuring-diesel-cli
+
+[print_schema]
+file = "fastn-core/src/schema.rs"
+custom_type_derives = ["diesel::query_builder::QueryId"]
+
+[migrations_directory]
+dir = "migrations"
```
amitu, line 9: Cool! I think they can take care of migration stuff for us.

fastn-core/Cargo.toml
```diff
@@ -80,6 +80,8 @@ fastn-observer.workspace = true
 fastn-js.workspace = true
 argon2.workspace = true
 uuid.workspace = true
+lettre.workspace = true
```
amitu, line 83: I think this is a decent idea as well. We will soon have a sort of "fastn standard environment", which will provide PostgreSQL, Redis, and SMTP.

migrations/2023-12-18-114245_fastn_auth_tables/down.sql
```diff
@@ -0,0 +1,11 @@
+-- This file should undo anything in `up.sql`
+
+DROP TABLE IF EXISTS fastn_email_confirmation;
+DROP TABLE IF EXISTS fastn_user_email;
+DROP TABLE IF EXISTS fastn_oauthtoken;
+DROP TABLE IF EXISTS fastn_session;
+DROP TABLE IF EXISTS fastn_user;
+
+DROP DOMAIN IF EXISTS email;
+DROP EXTENSION IF EXISTS citext;
```
amitu, line 10: Checkout the TIP in https://www.postgresql.org/docs/current/citext.html#id-1.11.7.19.9:

> Consider using nondeterministic collations (see [Section 24.2.2.4](https://www.postgresql.org/docs/current/collation.html#COLLATION-NONDETERMINISTIC)) instead of this module.

They are discouraging using of `citext` now, but I am not sure what exactly they are recommending, like how to apply their tip in concrete manner. Do read it up and let me know if it makes sense to you.

amitu:
@siddhantk232 also checkout https://github.com/orgs/fastn-stack/discussions/1449, and start thinking of ways we can make progress there.

diesel.toml
siddhantk232: yeah, I was thinking of wrapping their migration feature for `fastn migrate` command :smiling_imp:

fastn-core/src/auth/emailpassword.rs
Arpita-Jaiswal, line 1: Don't use `use` instead give the whole path.

diesel.toml
amitu: This will be awesome.

Cargo.toml
```diff
@@ -145,6 +145,8 @@ zip = "0.6"
 prettify-js = "0.1.0"
 indexmap = { version = "2", features = ["serde"] }
 argon2 = "0.5.2"
+lettre = { version = "0.11.2", features = ["serde", "tokio1", "tokio1-native-tls"]}
+diesel = { version = "2.1.4", features = ["postgres_backend", "r2d2"]}
```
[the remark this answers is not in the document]
siddhantk232: added in [34997bf](https://github.com/fastn-stack/fastn/pull/1612/commits/34997bfa48adfec03f834d914c6e2aebb27a71b8)

fastn-core/src/auth/emailpassword.rs
```diff
@@ -0,0 +1,371 @@
...
+
+    async fn pool() -> &'static Result<deadpool_postgres::Pool, deadpool_postgres::CreatePoolError>
+    {
+        POOL_RESULT.get_or_init(create_pool).await
+    }
+
+    pub async fn get_client() -> ftd::interpreter::Result<deadpool_postgres::Object> {
+        let _lock = EXECUTE_QUERY_LOCK.lock().await;
+
+        let client = pool().await.as_ref().unwrap().get().await.unwrap();
```
Arpita-Jaiswal, line 367: Convert this to appropriate result. Remove unwrap.

fastn-core/src/auth/github/mod.rs
```diff
@@ -67,26 +72,44 @@ pub async fn callback(
...
-    let user_detail_str = serde_json::to_string(&ud)?;
+    let user = fastn_core::auth::emailpassword::upsert_user(
+        &user_id,
+        gh_user
+            .email
+            .expect("email must exist for github user")
+            .as_str(),
+        gh_user.login.as_str(),
+        // TODO: should present an onabording form that asks for a name if github name is null
+        gh_user.name.unwrap_or(String::new()).as_str(),
```
Arpita-Jaiswal, line 85:
```suggestion
        gh_user.name.unwrap_or_default().as_str(),
```

fastn-core/src/auth/mod.rs
siddhantk232: I have made changes. Now these are not optional and `id` is not `uuid`.

fastn-core/src/auth/email_password.rs
```diff
@@ -0,0 +1,166 @@
...
+        &argon2::Argon2::default(),
+        payload.password.as_bytes(),
+        &parsed_hash,
+    );
+
+    if password_match.is_err() {
+        return fastn_core::http::api_error(
+            "incorrect username/password",
+            fastn_core::http::StatusCode::BAD_REQUEST.into(),
+        );
```
amitu, line 151: The error handling is still not correct.

fastn-core/src/auth/emailpassword.rs
siddhantk232: `fastn_core::db` exists now and used by the auth code. We'll incorporate using this in the `sql` processor in a separate pr.

fastn-core/src/auth/email_password.rs
amitu: We have to return something that fastn can consume: https://github.com/fastn-stack/fastn/blob/main/fastn-js/js/ftd.js#L218C45-L229

migrations/2023-12-19-114249_fastn_auth_remove_nullables/up.sql
```diff
@@ -0,0 +1,25 @@
+-- Your SQL goes here
```
amitu, line 1: For now let's coalesce all migrations into a single migration. When this PR gets merged and the first release is created we should have only two migrations, one that diesel creates for its own purpose, and the other that will initialise all tables etc.

siddhantk232: Sure, I'll do that.

I was thinking of using [diesel_migrations](https://docs.rs/crate/diesel_migrations/2.0.0)'s `embed_migration!` to pack and run these migrations on `FASTN_DB_URL` if say `FASTN_ENABLE_AUTH=true`. What do you think?

Later when we have `fastn migration` subcommand, we can generate migrations for auth related tables in user's project (like django) if they've enabled the auth feature.

fastn-core/src/auth/email_password.rs
siddhantk232: There are bunch of compilation errors that are still need to be resolved (related to gh auth and the `user-details` processor). This PR is still a WIP. I'll definietly write proper json responses for error states once those errors are resolved.

Tho I think, ["incorrect username/password" is ok](https://superuser.com/a/1048752) as we do not want to let the client know what exactly went wrong, do we?

migrations/2023-12-19-114249_fastn_auth_remove_nullables/up.sql
amitu: Our migrations should be embedded in fastn itself. `fastn migrate` is a tool so end users can create their own tables and manage migrations. They are technically orthogonal, we can manage our tables using a different mechanism that how we manage end user's tables.

But I think diesel can take care of both. One set of migration will have to be embedded and other set discovered (and not just in top level package, but a all packages that have db access, how do we give db access is not yet decided).

fastn-core/src/http.rs
```diff
@@ -687,3 +686,81 @@ pub async fn github_graphql<T: serde::de::DeserializeOwned>(
...
+/// ```json
+/// {
+///     data: null,
+///     errors: {
+///         key: Vec<String>
+///     }
+/// }
+/// ```
+pub async fn user_err(
+    errors: Vec<(&str, Vec<&str>)>,
```
amitu, line 701: The code in `ftd.js` handles both `errors: Vec<(&str, Vec<&str>)>,` and `errors: Vec<(&str, &str)>,`, and I think we never send a list of errors for same keys, its a provision django has, but I have never used. So lets simplify our lives by seeing single string for each key.

siddhantk232: [e90639](https://github.com/fastn-stack/fastn/pull/1612/commits/e90639e9666b3b5cc867d590efb56e82afc6e846)

fastn-core/src/auth/email_password.rs
```diff
@@ -164,3 +202,24 @@ pub(crate) async fn login(
     // https://stackoverflow.com/a/39739894
     fastn_core::auth::set_session_cookie_and_end_response(req, session_id, next).await
 }
+
+fn confirmation_mail_body(link: String) -> String {
+    format!("Use this link to verify your email: {link}")
+}
```
amitu, line 208: Store this in an external file so end user can modify it. Do a fallback in case the file does not exit.

diesel.toml
siddhantk232: This will be addressed in a separate PR

fastn-core/src/auth/mod.rs
siddhantk232: tests using `fastn test` will be added in a separate PR for this auth feature.

Cargo.toml
```diff
@@ -144,6 +144,20 @@ winit = "0.29"
 zip = "0.6"
 prettify-js = "0.1.0"
 indexmap = { version = "2", features = ["serde"] }
+argon2 = "0.5.2"
```
amitu, line 147: Do not use precise versioning unless you really care. Prefer `0.5` over `0.5` and `1` over `1.3.5` etc.

fastn-core/src/auth/mod.rs
```diff
@@ -27,58 +40,74 @@ impl AuthProviders {
...
+                        fastn_core::auth::get_authenticated_user_with_email(&session_id).await?;
+
+                    match requested_field {
+                        "username" | "user_name" | "user-name" => Ok(Some(user.username)),
+
+                        "token" => {
+                            use diesel::prelude::*;
+                            use diesel_async::RunQueryDsl;
+
+                            let pool = fastn_core::db::pool().await.as_ref().unwrap();
```
amitu, line 70: Don't `unwrap()`.

migrations/2023-12-18-114245_fastn_auth_tables/down.sql
siddhantk232: The `COLLATE` features seems nice and they've provided some examples too: (including case-insensitive comparisons) [https://www.postgresql.org/docs/current/collation.html#COLLATION-NONDETERMINISTIC](postgresql.org/docs/current/collation.html#COLLATION-NONDETERMINISTIC)

It seems complex to implement than the current `citext` we already have. I am going to ignore this for now as I don't fully understand the consequences of locale handling (I doubt there are any in this case)

fastn-core/src/auth/email_password.rs
siddhantk232: This will be done in another PR :+1:

fastn-core/src/auth/emailpassword.rs
siddhantk232: Are you sure? This `POST` requests will be done by clients (`fetch` for example) and they'll have to [handle the redirect anyways](https://stackoverflow.com/a/39739894), shouldn't we just send the json response and let them handle the redirection?

multiauth.sql
siddhantk232: the `updated_at` field had been added.

The current github oauth flow does not receive refresh_tokens. I'll change this to get refresh_token and store it in `fastn_oauthtoken` table along with logic to refresh tokens in another PR

fastn-core/src/auth/mod.rs
```diff
@@ -110,3 +176,90 @@ pub async fn get_auth_identities(
...
+    Ok((user, email.0))
+}
+
+pub(crate) fn enable_auth() -> fastn_core::Result<()> {
+    use diesel::Connection;
+    use diesel_migrations::MigrationHarness;
+    let db_url = std::env::var("FASTN_DB_URL")?;
+
+    // change in migrations require a change in rust code for to run
+    // see the docs of this function for the workaround that we do not need right now
```
amitu, line 249: Instead of these, let's add a `build.rs`:

```rs
fn main() {
    println!("cargo:rerun-if-changed=migrations");
}
```

amitu (approved):

siddhantk232:
a demo created using changes introduced in this PR: https://github.com/siddhantk232/fastn-test-multiauth

@amitu I think we can merge this PR now. Remaining things will be done in a new PR

amitu (approved):

fastn-core/src/auth/emailpassword.rs
amitu: But you are not sending JSON with redirection info, you are sending user data serialised as JSON. What will client do with non redirect json?

PR #1645, opened 2023-12-27
take record args for fetch options in ftd.fetch

siddhantk232:
### Example

```
-- ftd.http-options opts:
redirect: manual ;; only follow | manual | error     default: manual
method: GET  ;; only GET | POST                      default: GET
fastn-module: test-multiauth

-- ftd.column:
padding.px: 16

-- ftd.text: Login
padding-top.px: 16
$on-click$: $submit(opts = $opts)

-- end: ftd.column

-- void submit(opts):
ftd.http-options opts:

ftd.http(
    "/-/auth/login/?provider=emailpassword",
    opts,
    ("username", "test"),
    ("password", "example"),
)
```


amitu:
How did you format the code? Maybe we should add formatter as CI, the way we do for Rust to achieve consistent formatting of JS files.

amitu:
We have to move the definition of the record into ftd crate, with name `ftd.http-options` or something.

amitu:
Instead of `string redirect` we should have an `or-type`, `ftd.http-redirect-behaviour` or something.

amitu:
Method should also be an `or-type`.

siddhantk232:
> How did you format the code? Maybe we should add formatter as CI, the way we do for Rust to achieve consistent formatting of JS files.

```js
prettier --write --tab-width 4 fastn-js/js/ftd.js
```

amitu (approved):

PR #1650, opened 2023-12-28
Test

harshdoesdev:


harshdoesdev:
🤦  Was trying to integrate JS formatting using prettier, please ignore this

amitu:
@harshdoesdev Ah, why did you stop?! I want this :-). Ideally it should be part of our rust.yml (which we should rename to `tests.yml`.

harshdoesdev:
> @harshdoesdev Ah, why did you stop?! I want this :-). Ideally it should be part of our rust.yml (which we should rename to `tests.yml`.

Cool! here is the PR: https://github.com/fastn-stack/fastn/pull/1654

PR #1655, opened 2023-12-29
more fastn test updates

Heulitig:


.github/workflows/tests_and_formatting.yml
```diff
@@ -74,6 +74,11 @@ jobs:
       - name: Check if test succeeded
         if: steps.test.outcome != 'success'
         run: exit 1
+      - name: Check out
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.head_ref }}
+          fetch-depth: 0
```
amitu, line 81: Why have you added this? We already have another checkout step at line number 14.

harshdoesdev: The "prettier-action" requires fetch-depth: 0 to work with the only_changed: True option enabled. Fetching with a depth of 0 is resource-intensive, which is why I am executing it only after all Rust checks have passed.

amitu: Make fetch-depth 0 at the top. You are trying to make it fast I understand, but I hope you understand that your end result is making it slow.

harshdoesdev: Got it. I am creating a separate PR for this fix.

amitu: Also fetch-depth 0 should be used if you are trying to be smart and only run prettier on files that have modified, it does not make sense for us, we have hardly ~1K lines of code to check, no matter how slow prettier is, it will be faster to recheck all files instead of downloading the entire git history just so we can find what changed.

amitu: Also use https://github.com/dprint/check, its written in Rust, will probably be faster to install (only a binary) and run.

fastn/src/main.rs
```diff
@@ -388,6 +389,7 @@ fn app(version: &'static str) -> clap::Command {
                     .action(clap::ArgAction::Append))
                 .arg(clap::arg!(--edition <EDITION> "The FTD edition"))
                 .arg(clap::arg!(--"script" "Generates a script file (for debugging purposes)"))
```
amitu, line 391: What is the script file?

Arpita-Jaiswal (approved):

PR #1664, opened 2024-01-02
fastn update

harshdoesdev:


amitu:
@harshdoesdev why are you creating a zip file? Github will create the zip file. `fastn update` will NOT create any zip file, only create a `manifest.json` and `config.json` files.

fastn-update/src/lib.rs
```diff
@@ -40,11 +40,28 @@ impl Manifest {
...
+}
+
+impl PackageSource {
+    pub fn github(username: String, repository: String) -> Self {
+        PackageSource::Github {
+            username,
+            repository,
+        }
+    }
+}
```
amitu, line 58: No, it should just contain a zip url, you do not need all this logic.

fastn-update/src/lib.rs
```diff
@@ -2,142 +2,82 @@ use std::io::Read;
...
+    config: &fastn_core::Config,
+    packages_root: &camino::Utf8PathBuf,
+    dependencies: &Vec<fastn_core::package::dependency::Dependency>,
+    packages: &mut Vec<fastn_update::types::Package>,
+) -> fastn_core::Result<()> {
+    for dependency in dependencies {
+        match fastn_update::utils::get_package_source_url(&dependency.package) {
+            Some(source) => {
+                let (mut archive, checksum) =
+                    fastn_update::utils::download_archive_from_gh(source.clone()).await?;
```
amitu: This should not be `github` specific. It must work as long as you have a zip file URL available. The fact that we are using Github or Github is generating should not be your concern/assumption.

harshdoesdev: yes, I have refactored this in the latest commit

amitu:
@harshdoesdev the location of `manifest.json` is going to be inside `.build` (or .fastn in future, but since we do not yet have it lets use `.build`). Also for each file you have to store file name, file size, and hash. Which means you can not store manifest itself in the manifest (how will you generate the hash?).

fastn-core/src/commands/build.rs
```diff
@@ -19,6 +21,47 @@ pub async fn build(
...
+            let _ = &config
+                .ds
+                .write_content(
+                    &build_dir.join("manifest.json"),
+                    serde_json::ser::to_vec_pretty(&manifest)?,
+                )
+                .await?;
+
+            println!("Wrote manifest.json");
+        }
```
Arpita-Jaiswal, line 63: Move this inside a function.

fastn-core/src/commands/build.rs
```diff
@@ -19,6 +21,47 @@ pub async fn build(
     {
         let documents = get_documents_for_current_package(config).await?;
 
+        if let Some(zip_url) = zip_url {
+            use sha2::digest::FixedOutput;
+            use sha2::Digest;
+
```
Arpita-Jaiswal, line 27: Add `println!("Processing manifest.json ...")` somewhat same as for other files.

fastn-core/src/commands/build.rs
```diff
@@ -19,6 +21,47 @@ pub async fn build(
...
+            let _ = &config
+                .ds
+                .write_content(
+                    &build_dir.join("manifest.json"),
+                    serde_json::ser::to_vec_pretty(&manifest)?,
+                )
+                .await?;
+
+            println!("Wrote manifest.json");
```
Arpita-Jaiswal, line 62: Check what we write for other files.

Heulitig (approved):

Heulitig (approved):

PR #1665, opened 2024-01-02
WIP DON'T MERGE

siddhantk232:


Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ekidd/rust-musl-builder
```
amitu, line 1: We should move this to `.github` folder.

PR #1669, opened 2024-01-02
testing integration tests

Heulitig:


.github/workflows/integration_testing.yml
```diff
@@ -80,3 +81,82 @@ jobs:
           for folder in integration-tests/*/; do
             (cd "$folder" && fastn test --headless)
           done
```
amitu: This is wrong, `fastn test` should do the equivalent of this internally.

.github/workflows/integration_testing.yml
```diff
@@ -80,3 +81,82 @@ jobs:
...
+        run: |
+          timeout=30
+          until pg_isready -h localhost -p 5432 -U testuser -d testdb || ((timeout-- <= 0)); do
+            sleep 1
+          done
+
+      - name: Install pip dependencies (sourcing auto.sh)
+        id: navigate-inside-slides
+        run: |
+          source scripts/auto.sh
```
amitu, line 141: Does this work? I mean environment variables created in one step visible to other steps?

.github/workflows/integration_testing.yml
```diff
@@ -80,3 +81,82 @@ jobs:
...
+          # This will need secret key
+          python backend/manage.py runserver 7999 &
+          
+          # Waiting for the server to start
+          sleep 10  # We might want to wait more
+
+      - name: Run integration tests
+        run: |
+          cd frontend
+          fastn test --headless
```
amitu, line 163: Maybe we can combine this by passing `--path=frontend` to `fastn test`.

.github/workflows/integration_testing.yml
```diff
@@ -80,3 +81,82 @@ jobs:
           for folder in integration-tests/*/; do
             (cd "$folder" && fastn test --headless)
           done
+
+  slides-integration-test:
```
amitu, line 85: Why is this called slides test? Slides related tests should be in slides repo. Integration test in fastn should test fastn itself, the auth feature, endpoint using some example project that is part of this repo just for testing.

.github/workflows/integration_testing.yml
```diff
@@ -80,3 +81,82 @@ jobs:
...
+    services:
+      postgres:
+        image: postgres:latest
+        env:
+          POSTGRES_DB: fifthtry
+          POSTGRES_USER: admin
+          POSTGRES_PASSWORD: admin
+        ports:
+          - 5432:5432
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
```
amitu, line 101: I think we can reduce the interval to make this step faster. There is no harm is testing every second or even more frequently, it is our action, machine is not doing anything else etc.

.github/workflows/integration_testing.yml
```diff
@@ -80,3 +81,82 @@ jobs:
...
+          profile: minimal
+          override: true
+          components: rustfmt, clippy
+
+      - name: fastn installation (using source)
+        id: fastn-install
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: --path fastn --profile dev
```
amitu, line 119: Why profile dev? Release would run the test faster.

.github/workflows/integration_testing.yml
```diff
@@ -80,3 +81,82 @@ jobs:
...
+
+      - name: Clone Slides repository
+        id: clone-slides-repo
+        run: |
+          git clone https://github.com/FifthTry/slides.git
```
amitu, line 124: This is wrong.

integration-tests/hello-world-sql.ftd
```diff
@@ -0,0 +1,13 @@
+-- import: fastn/processors as pr
```
amitu, line 1: Add a similar http processor test.

scripts/hello-world-test.py
```diff
@@ -0,0 +1,40 @@
+import psycopg2
```
amitu, line 1: Delete this file,.

.github/workflows/integration_testing.yml
```diff
@@ -0,0 +1,81 @@
...
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          override: true
+          components: rustfmt, clippy
```
amitu, line 39: Do we really need this step?

.github/workflows/integration_testing.yml
```diff
@@ -0,0 +1,81 @@
...
+
+      - name: fastn installation (using source)
+        id: fastn-install
```
amitu, line 42: Remove the id.

.github/workflows/integration_testing.yml
```diff
@@ -0,0 +1,81 @@
...
+        continue-on-error: true
```
amitu, line 43: Remove this line.

.github/workflows/integration_testing.yml
```diff
@@ -0,0 +1,81 @@
...
+        run: |
+          python scripts/test-server.py &
+          
+          # Waiting for the server to start
+          sleep 10  # We might want to wait more
+
+      - name: Run integration tests
+        run: |
+          cd integration-tests
+          fastn test --headless
```
amitu, line 81: Move all these to a single script: `scripts/run-intergration-test.sh`. Set -e in the script so we see when it fails.

amitu: `set -eou pipefail`

amitu: `set -o xtrace`

amitu (approved):

PR #1670, opened 2024-01-03
use quick-js on non windows platforms

siddhantk232:
Building with `rquickjs` for linux and mac throws errors but this lib doesn't seem to work correctly at runtime. I have reverted the [changes to use `rquickjs`](https://github.com/fastn-stack/fastn/pull/1668) for now.

- `.dockerignore` to reduce docker image size

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,29 @@
+# editor and OS junk
+.idea
+**/.DS_Store
+
+ftd/t/js/**.manual.html
+ftd/t/js/**.script.html
+
+# nix symlink to the build output
+result
+
+.direnv
+.envrc
+
+# Rust stuff
+target
+**/*.rs.bk
+
+# fastn test
+fastn-core/tests/**/.build
+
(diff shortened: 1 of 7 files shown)


amitu:
@siddhantk232 can you do a version bump also, I will merge, create release and test on Heroku.

siddhantk232:
> @siddhantk232 can you do a version bump also, I will merge, create release and test on Heroku.

done @amitu

amitu:
Update the Cargo.lock file as well (it happens if you run `cargo check` etc).

siddhantk232:
> Update the Cargo.lock file as well (it happens if you run `cargo check` etc).

done :slightly_smiling_face:

siddhantk232:
Added a fix for the `user-details` processors as well. When logged out or when the `auth` feature is disabled, it'll give the default user details, i.e., logged out user.

@amitu

amitu (approved):

PR #1682, opened 2024-01-04
Interpreter Fix: Or-type component property

Arpita-Jaiswal:
Fixes https://github.com/fastn-stack/fastn/issues/1681


amitu:
@Arpita-Jaiswal this fixes the linked issue, but ui.fifthtry.com is still not building:

image

PR #1683, opened 2024-01-04
integration testing changes

Heulitig:


scripts/run-integration-tests.sh
```diff
@@ -0,0 +1,29 @@
...
+
+echo "Installing python server dependencies"
+python -m pip install --upgrade pip
+pip install Flask psycopg2
+
+echo "Waiting for postgres to be ready"
+timeout=30
+until pg_isready -h localhost -p 5432 -U testuser -d testdb || ((timeout-- <= 0)); do
+  sleep 1
+done
```
amitu, line 17: Add a command line argument (pass it on CI, on local no need to add argument).

scripts/run-integration-tests.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+export FASTN_ROOT=`pwd`
+export FASTN_DB_URL=postgres://testuser:testpassword@localhost:5432/testdb
```
amitu, line 3: Put this also only if `--ci` argument is supplied.

amitu, line 3: `export FASTN_DB_URL=${FASTN_DB_URL-postgres://testuser:testpassword@localhost:5432/testdb}` (`-` means if what's on left is not defined, use what's on right, do confirm it somewhere).

scripts/run-integration-tests.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Enable xtrace and pipefail
+set -o xtrace
+set -eou pipefail
+
+echo "Installing python server dependencies"
+python -m pip install --upgrade pip
```
amitu, line 10: We should have .venv for local so we do not corrupt global python.

amitu (approved):

PR #1696, opened 2024-01-09
fastn update

harshdoesdev:


fastn-update/src/lib.rs
```diff
@@ -2,60 +2,91 @@ extern crate self as fastn_update;
...
+        #[source]
+        error: fastn_core::Error,
+    },
+    #[error("Failed to resolve dependency '{package}'")]
+    DependencyResolution {
+        package: String,
+        #[source]
+        error: fastn_core::Error,
+    },
+}
```
amitu, line 25: Don't use this pattern for new code. Create a unique error enum for every function called with precise error variants.

fastn-update/src/lib.rs
```diff
@@ -2,60 +2,91 @@ extern crate self as fastn_update;
...
+
+type Result<T> = std::result::Result<T, Error>;
```
amitu, line 27: Since we have a lot of Error types, we do not need Result alias.

fastn-update/src/lib.rs
```diff
@@ -2,60 +2,91 @@ extern crate self as fastn_update;
...
+
 async fn resolve_dependencies(
     ds: &fastn_ds::DocumentStore,
     packages_root: &fastn_ds::Path,
     current_package: &fastn_core::Package,
-) -> fastn_core::Result<()> {
+    pb: &indicatif::ProgressBar,
```
amitu, line 33: 👍

Arpita-Jaiswal (approved):

PR #1706, opened 2024-01-12
adding multi endpoint 

Heulitig:


integration-tests/FASTN.ftd
```diff
@@ -1,4 +1,9 @@
 -- import: fastn
 
 -- fastn.package: integration-tests
-endpoint: http://127.0.0.1:5000
+
+-- fastn.endpoint: http://fastn.com
+mountpoint: /ftd/
```
amitu, line 6: Let's revers the order, I think this would look better:

```ftd
-- fastn.package: foo.com

-- fastn.mount: /ftd/
forward-to: http://fastn.com
include-prefix: true ;; true is default
```

If `include-prefix` is `true`, `foo.com/ftd/hello` will be forwarded to `http://fastn.com/ftd/hello`, if it was `false`, it will be forwarded to `http://fastn.com/hello`.

amitu: Or should we put it inside `fastn.url-mappings`](https://github.com/orgs/fastn-stack/discussions/1623)?

```
-- fastn.url-mappings:

/ftd/* -> http://fastn.com/ftd/*
```

In this case how would we differentiate from redirection?

amitu: Maybe:

```
-- fastn.url-mappings:

/ftd/* -> http+proxy://fastn.com/ftd/*
```

fastn-core/src/package/redirects.rs
```diff
@@ -1,18 +1,84 @@
...
+    pub(crate) fn url_mappings_from_body(&self) -> fastn_core::Result<UrlMappings> {
+        let url_mappings_body = self.body.as_str();
+        let redirects = self.find_redirects(url_mappings_body)?;
+        let endpoints = self.find_endpoints(url_mappings_body)?;
+
+        let mut url_mappings = UrlMappings::new();
+        url_mappings.redirects = redirects;
+        url_mappings.endpoints = endpoints;
+        Ok(url_mappings)
+    }
```
amitu: This is a good function to unit test. Write a few.

integration-tests/FASTN.ftd
```diff
@@ -5,4 +5,4 @@
 -- fastn.url-mappings:
 
 /ftd/* -> http+proxy://fastn.com/ftd/*
-/get-data/ -> localhost+proxy:7999/get-data/
+/test-server-data/ -> localhost+proxy:5000/get-data/
```
amitu, line 8: What is localhost+proxy? This is wrong. It should be `http+proxy://localhost:5000/get-data/`. Also we should not allow proxy without `*`, so this line must be `/test-server-data/* -> localhost+proxy:5000/get-data/*`.

integration-tests/FASTN.ftd
```diff
@@ -1,4 +1,8 @@
 -- import: fastn
 
 -- fastn.package: integration-tests
-endpoint: http://127.0.0.1:5000
+
+-- fastn.url-mappings:
+
+/ftd/* -> http+proxy://fastn.com/ftd/*
+/test-server-data/* -> http+proxy://localhost:5000/get-data/*
```
amitu, line 8: Add a redirect test also.

amitu (approved):

PR #1708, opened 2024-01-16
Feat/test auth routes

siddhantk232:
This PR changes the existing `fastn.test` framework to make it work with the added integration tests for the authentication related code in `fastn_core::auth`.

- moved route handlers to the `fastn_core::commands::serve::serve_helper` which is used by the `fastn_core::commands::serve::serve` and the testing framework's `fastn.get` and `fast.post` handlers.

## Auth Code change:

- Fix the `send_raw` method in `mail.rs` file to only construct message using [lettre](https://docs.rs/lettre/latest/lettre/) when `DEBUG!=true`
- send test data as `X-Fastn-Test-` headers in the signup (create_user) route handler
- handle the case in login where the user tries to login before verifying their email.

I don't expect the integration tests added in this PR to work in CI as we might need some env variables to trigger the db migration and a fastn built with the `auth` feature flag. But we can fix that in a later PR where I add the [remaining tests](https://github.com/fastn-stack/fastn/pull/1708/files#diff-b199dbb694cdbfc1f58506650f46a61d158e135705ee870c814e4f1773fc885dR78)


siddhantk232:
> > the existing `testn` framework inside `fastn`
> 
> `testn` was not meant for `fastn test`, it was meant for a binary we will create to facilitate testing ci environments, like github action, for people who store their ftd code on FifthTry. We have created a repo [`FifthTry/clift`](https://github.com/FifthTry/clift) and are not using `testn` name, as much as I like :-). fastn and FifthTry are two brands already, adding more would be tricky.

Thanks! I have changed the description

fastn-core/src/auth/mod.rs
```diff
@@ -56,8 +58,34 @@ pub async fn get_user_data_from_cookies(
...
+
+                            if e == AuthUserError::UserDoesNotExist {
+                                return Err(fastn_core::Error::GenericError(
+                                    "User does not exist".to_string(),
+                                ));
+                            } else if let AuthUserError::UserExistsWithUnverifiedEmail(_) = e {
+                                return Err(fastn_core::Error::GenericError(
+                                    "User is not verified".to_string(),
+                                ));
+                            }
```
amitu, line 80: These errors are not compatible with out `ftd.js`'s `http()` function.

amitu: Write the code with the assumption that only http() function would be calling all these things. Nothing else is supported (we will never be "proper api" for outsiders to integrate).

integration-tests/_tests/12-auth-create-user.test.ftd
```diff
@@ -0,0 +1,85 @@
...
+
+
+-- fastn.post: login with valid but unverfied user
+url: /-/auth/login/?provider=emailpassword&next=/hello/
+body: {"username": "john", "password": "testpasswd"}
+http-status: 400
+
+-- fastn.post.test:
+
+fastn.assert.eq(fastn.http_response.errors.username, "User is not verified");
```
amitu, line 75: Seems they are compatible with it. My bad.

integration-tests/_tests/12-auth-create-user.test.ftd
```diff
@@ -0,0 +1,85 @@
+;; require fastn built using --features auth flag
+;; FASTN_DB_URL, FASTN_ENABLE_AUTH must be set
+-- import: fastn
+
+-- fastn.test: create user
+
+-- fastn.post: Create user with insufficient data should fail
+url: /-/auth/create-user/
+body: {"username": "siddhant"}
+http-status: 400
```
amitu, line 10: Do we handle 400 status code in `ftd.http()`?

integration-tests/_tests/12-auth-create-user.test.ftd
```diff
@@ -0,0 +1,85 @@
+;; require fastn built using --features auth flag
+;; FASTN_DB_URL, FASTN_ENABLE_AUTH must be set
+-- import: fastn
+
+-- fastn.test: create user
+
+-- fastn.post: Create user with insufficient data should fail
```
amitu, line 7: I think we should rename these to `fastn.view` and `fastn.action`. Also for actions that are failing we can create `fastn.failing-action`, so we do not have to explicitly assert error code etc. We can even assert error messages:

```ftd
-- fastn.failing-action: Create user with insufficient data should fail
url: /-/auth/create-user/
email-error: Email is required

{
    "username": "siddhant"
}
```

Or even better:

-- fastn.failing-action: Create user with insufficient data should fail
url: /-/auth/create-user/
username: siddhant
email-error: Email is required
```

Comment (amitu) [reply integration-tests/_tests/12-auth-create-user.test.ftd:7]: We can leave `fastn.get` and `fastn.post` and these new helpers `fastn.view`, etc.

Comment (amitu) [reply integration-tests/_tests/12-auth-create-user.test.ftd:7]: In another PR.

File: integration-tests/_tests/12-auth-create-user.test.ftd (hunk 3 above)
Comment (siddhantk232) [reply integration-tests/_tests/12-auth-create-user.test.ftd:10]: we don't at the moment. I can do a quick change in `ftd.js` to handle this if you want me to

Comment (amitu) [reply integration-tests/_tests/12-auth-create-user.test.ftd:10]: No, we should return 200 error code. This kind of debate about what is the right error code is bike shedding. Would you return 400 for the case when the username entered does not exist? You monster! Not founds are 404! Some some sort of form field errors are 404 and some 400? And what if someone is not allowed to edit a document, would we return 400 or 401? If username is taken, is it same as "for example, malformed request syntax, invalid request message framing, or deceptive request routing" and both should receive same error code, if one did not care of error code, it would have been excusable, but someone claiming to be error code purist doing so? Not so much.

200 for proper scenarios. 4xx for when something is seriously wrong with the client. Form handling is not something is wrong with client. Everything is perfectly fine with client. User data error is not "client error".

Comment (siddhantk232) [reply integration-tests/_tests/12-auth-create-user.test.ftd]: Thanks for explaining. This makes sense. I have [changed user errors to be 200 OK](https://github.com/fastn-stack/fastn/pull/1708/commits/62da38dccb170a7058cafe21b5b439ad74f8b013)

File: integration-tests/_tests/12-auth-create-user.test.ftd [hunk 5]
```diff
@@ -48,7 +48,7 @@ fastn.assert.not_empty(fastn.http_response.errors.payload);
 -- fastn.post: login with invalid user shouldn't work
 url: /-/auth/login/?provider=emailpassword
 body: {"username": "someone", "password": "veryrealpasswd"}
-http-status: 400
+http-status: 200
```
amitu: This should be default, so let's leave this line out. (do check if fastn.test def asserts http code 200 even if the http-status is not explicitly passed).

amitu (approved):

PR #1710, opened 2024-01-17
move migrations to run in the serve_helper and fix for the static file handler

siddhantk232:
- Return `None` when static_file [fails](https://github.com/fastn-stack/fastn/pull/1710/files#diff-5fb4b883ae3e728c7451b8b0f17acd1d8560caf2fb85a2b9484d840aaa93e824R556) so that the serve_helper can take over and try serving the static files through the `endpoint:` set in package's FASTN.ftd. This is helpful in situations where the endpoint also wants to serve static routes (routes ending with .css for example). For example, django admin if used as an endpoint requires some css files that are served by the django server.
- Moving migrations in serve_helper adds ~15ms to each request which is really bad but this is possible the simplest thing I could come up with so that these migrations are run for `fastn test`, `fastn server` and outside when `fastn-core` crate is used as a dependency (in hostn)


fastn-core/src/commands/serve.rs
```diff
@@ -191,6 +191,15 @@ pub async fn serve_helper(
...
     only_js: bool,
 ) -> fastn_core::Result<fastn_core::http::Response> {
+    #[cfg(feature = "auth")]
+    if let Ok(auth_enabled) = std::env::var("FASTN_ENABLE_AUTH") {
+        // only apply migrations when FASTN_ENABLE_AUTH=true
+        if auth_enabled == "true" {
+            tracing::info!("running auth related migrations");
+            fastn_core::commands::serve::enable_auth()?
+        }
+    }
```
amitu, line 201: Problem is this gets called on every HTTP request, and we do not want to run migration on every HTTP request. We have to explicitly call this method from `hostn` when `fastn_ds::DocumentStore()` is getting constructed for any domain.

siddhantk232:
> * Return `None` when static_file [fails](https://github.com/fastn-stack/fastn/pull/1710/files#diff-5fb4b883ae3e728c7451b8b0f17acd1d8560caf2fb85a2b9484d840aaa93e824R556) so that the serve_helper can take over and try serving the static files through the `endpoint:` set in package's FASTN.ftd. This is helpful in situations where the endpoint also wants to serve static routes (routes ending with .css for example). For example, django admin if used as an endpoint requires some css files that are served by the django server.
> * Moving migrations in serve_helper adds ~15ms to each request which is really bad but this is possible the simplest thing I could come up with so that these migrations are run for `fastn test`, `fastn server` and outside when `fastn-core` crate is used as a dependency (in hostn)

1. Return `None` fix was made to let the `endpoint` handle the static files if we fail to do so. But this is not an intended behaviour. [Endpoints take precedence over static files already]() so this situation wouldn't exist to handle.

2. [As mentioned here](https://github.com/fastn-stack/fastn/pull/1710#discussion_r1455514127). We intend to have this be called in hostn outside of the `serve()` function and in fastn, it'll be called when the binary is run (for `fastn serve`, `fastn test` and others)

This leaves nothing in this PR so I'm closing this one :smile:

PR #1718, opened 2024-01-19
rename enable_auth -> migrate...

siddhantk232:
Rename `enable_auth()` to `fastn_core::db::migrate()` and prepare it for
external use


fastn-core/src/db.rs
```diff
@@ -16,3 +16,25 @@ static POOL_RESULT: tokio::sync::OnceCell<fastn_core::Result<PgPool>> =
 pub async fn pool() -> &'static fastn_core::Result<PgPool> {
     POOL_RESULT.get_or_init(create_pool).await
 }
+
+/// run migrations on `db_url`
+pub fn migrate(db_url: String) -> fastn_core::Result<()> {
```
amitu, line 21: Not too happy with it being sync operation. It can starve all our threads on `hostn`, and lead to DOS. We should convert it to async by using [cpupool](https://docs.rs/async-cpupool/latest/async_cpupool/).

Also there is this create: https://lib.rs/crates/diesel_async_migrations, means async migration should be "natively" possible with diesel.

amitu: I have tried diesel_async_migrations (maybe 20-30 mins only), and gave up, trait issues in diesel + async etc are complete nightmare. So cpupool is good enough for now.

siddhantk232: well it worked first try for me https://github.com/fastn-stack/fastn/pull/1718/commits/0f2b3d7cd2fa626992d475bc8d6019b106f40150 :eyes:

fastn-core/src/http.rs
```diff
@@ -735,7 +735,6 @@ pub fn user_err(
     }
 
     let resp = serde_json::json!({
-        "data": null,
         "errors": json_error,
     });
```
amitu, line 739: We should return a key `success: bool` so based on key we can easily detect if it was success or failure, else we have to do existence checks.

amitu (approved):

PR #1734, opened 2024-01-27
chore: update and fix fbt tests

harshdoesdev:
- Updated fbt tests for dependency alias
- Added fbt test for `--offline` flag of fastn build. We first update the packages and then run the fastn build command with `--offline` flag.


fastn-core/tests/19-offline-build/cmd.p1
```diff
@@ -0,0 +1,26 @@
+-- fbt:
+cmd: $FBT_CWD/../target/debug/fastn --test update && $FBT_CWD/../target/debug/fastn --test build --offline
+output: .build
+
+-- stdout:
+
+Updated 2 dependencies.
+Updated 2 dependencies.
```
amitu, line 8: Is this statement accurate? Do we really update 2 dependencies, or have we checked two dependencies here and found nothing to update. Update means change.

PR #1735, opened 2024-01-29
Refactor: Config generation and Dependency Resolution

harshdoesdev:
Refactoring how config is generated and dependency packages are handled.

Discussion: https://github.com/FifthTry/business/discussions/308


harshdoesdev:
Continued in: https://github.com/fastn-stack/fastn/pull/1736

PR #1736, opened 2024-01-29
Refactor: Config generation and Dependency Resolution

harshdoesdev:
Refactoring how config is generated and dependency packages are handled.

Discussion: https://github.com/FifthTry/business/discussions/308


fastn/src/main.rs
```diff
@@ -60,6 +60,9 @@ async fn fastn_core_commands(matches: &clap::ArgMatches) -> fastn_core::Result<(
         return fastn_core::create_package(name, path, download_base_url).await;
     }
 
+    let current_dir: camino::Utf8PathBuf = std::env::current_dir()?.canonicalize()?.try_into()?;
+    let ds = fastn_ds::DocumentStore::new(current_dir);
+
     let mut config = fastn_core::Config::read_current(true).await?;
```
amitu, line 66: You should not use read_current if you have a handle to ds.

fastn/src/main.rs
```diff
@@ -80,7 +83,7 @@ async fn fastn_core_commands(matches: &clap::ArgMatches) -> fastn_core::Result<(
         let inline_css = serve.values_of_("css");
         let offline = serve.get_flag("offline");
 
-        fastn_update::update(&config, offline).await?;
+        fastn_update::update(&ds, offline).await?;
```
amitu, line 86: Since config is not needed by fastn_update, we can re-order this function to not construct config before calling this.

harshdoesdev:
Closed in favour of https://github.com/fastn-stack/fastn/pull/1751

PR #1739, opened 2024-02-01
fix: checkbox state not being reflected in UI

harshdoesdev:
This fixes the following issues:
1. Value of `$CHECKED`, a special variable, was being correctly resolved, hence it was being read as undefined.
2. Checkbox's checked UI was not being updated.

Example:
```ftd
-- boolean $is-checked: false

-- ftd.row:
spacing.fixed.px: 5
color: $inherited.colors.text
align-content: center

-- ftd.checkbox:
checked: $is-checked
$on-click$: $ftd.set-bool($a = $is-checked, v = $CHECKED)

-- ftd.checkbox:
checked: $is-checked
$on-click$: $ftd.set-bool($a = $is-checked, v = $CHECKED)

-- end: ftd.row
```


amitu:
@harshdoesdev the PR has minor issue (JS code formatting related). ALso can you do a version bump so we can release this fix.

Arpita-Jaiswal (approved):

PR #1741, opened 2024-02-02
Auth/html email

siddhantk232:
Read the content of confirmation email from `<package>/email/confirmation-mail.html` file.


fastn-core/src/auth/email_password.rs
```diff
@@ -540,13 +542,17 @@ async fn create_and_send_confirmation_email(
         .first(&mut conn)
         .await?;
 
+    let file_path = fastn_ds::Path::new("email/confirmation-mail.html");
```
amitu, line 545: This has to be part of the auth package. So read it by ftd. We will store the content of the HTML in a ftd file. `auth.confirmation_mail_html` or something would be our "interface".

siddhantk232: done https://github.com/fastn-stack/fastn/pull/1741/commits/f4bbbcef39c4d70e09f6b1e22799bd2f3c0fc7b6

fastn-core/src/auth/email_password.rs
```diff
@@ -551,13 +546,55 @@ async fn create_and_send_confirmation_email(
...
+    // To use auth. The package has to have auto import with alias `auth` setup
+    let path = req_config
+        .config
+        .package
+        .eval_auto_import("auth")
+        .unwrap()
+        .to_owned();
+
+    let path = path
+        .strip_prefix(format!("{}/", req_config.config.package.name).as_str())
```
amitu, line 558: Nopes. This must be `auth.<something>`. The content of email will be part of auth package, not current package.

amitu: Create fastn-community.github.io/auth and switch to that to avoid confusion. We have the repo, just commit to it, why are you delaying it?

amitu: End user will only add some auth package as a dependency in their FASTN.ftd file and will not add any other file in their package to get auth working end to end.

PR #1744, opened 2024-02-02
fix: perform server-side rendering only when user-agent is a bot

harshdoesdev:
We should perform SSR only when the user-agent is a bot, otherwise we can skip the SSR step for normal users. It reduces the server response time from `3.7s` to `1.5s` (tested on fastn.com served locally).


fastn-core/src/http.rs
```diff
@@ -170,11 +170,17 @@ impl Request {
...
             .and_then(|v| v.to_str().ok())
             .and_then(|v| v.parse().ok())
     }
 
+    pub fn user_agent(&self) -> Option<String> {
+        self.headers
+            .get(actix_web::http::header::USER_AGENT)
+            .and_then(|v| v.to_str().map(|v| v.to_string()).ok())
+    }
+
```
amitu, line 183: Also add a is_bot() method here.

fastn-core/src/package/package_doc.rs
```diff
@@ -566,6 +566,10 @@ pub(crate) async fn read_ftd_2023(
     download_assets: bool,
     only_js: bool,
 ) -> fastn_core::Result<FTDResult> {
+    let is_bot = match config.request.user_agent() {
+        Some(user_agent) => KNOWN_BOTS.is_bot(&user_agent),
+        None => true,
+    };
```
amitu, line 572: Move this logic to `fastn_core::http::Request.is_bot()`.

fastn-core/src/package/package_doc.rs
```diff
@@ -621,10 +625,14 @@ pub(crate) async fn read_ftd_2023(
...
-            format!("{js_ftd_script}\n{js_document_script}").as_str(),
-        );
+        let ssr_body = if !is_bot {
+            EMPTY_HTML_BODY.to_string()
+        } else {
+            fastn_js::ssr_with_js_string(
+                &package_name,
+                format!("{js_ftd_script}\n{js_document_script}").as_str(),
+            )
+        };
```
amitu, line 635: Invert the logic.

amitu:
@harshdoesdev, write a test for this also somehow.

fastn-core/src/http.rs
```diff
@@ -788,6 +794,45 @@ pub fn user_err(
         .body(serde_json::to_string(&resp)?))
 }
 
+/// Checks whether a request was made by a Google/Bing bot based on its User-Agent
+/// Google crawlers and fetchers: https://developers.google.com/search/docs/crawling-indexing/overview-google-crawlers
+/// Bing crawlers: https://www.bing.com/webmasters/help/which-crawlers-does-bing-use-8c184ec0
+pub fn is_bot(user_agent: &str) -> bool {
```
amitu, line 813: Let's also include popular social media sites. We have to ensure they have access to social media graph meta data at least, not the full HTML, so if we can do that without doing full SSR that is also fine. @Arpita-Jaiswal is that possible?

amitu:
@harshdoesdev we have to make sure SSR happens when we are doing `fastn build`. Can you confirm that is the case?

amitu:
@harshdoesdev can we write some test to ensure bots are getting ssr, and non bots are not?

harshdoesdev:
> @harshdoesdev can we write some test to ensure bots are getting ssr, and non bots are not?

@amitu I attempted to create a test case for this using `fbt`, but it resulted in a stack overflow error when using `fastn serve`. Therefore, for the time being, can we incorporate this test directly into the workflow and write a shell or Python script for it?

PR #1750, opened 2024-02-05
chore: added tests for SSR

harshdoesdev:
This adds test for checking if SSR is being properly done for requests which have user agents of bots and for normal users SSR has to be skipped.


.github/workflows/test.yml
```diff
@@ -101,6 +101,7 @@ jobs:
         id: ssr-test
         continue-on-error: true
         run: |
+          cargo install --path=fastn
```
harshdoesdev, line 104: it looks like serve using debug build of fastn is giving `actix-web stack overflow` error, this is just to check if that is the case.

harshdoesdev: @amitu the tests are passing, but this step is taking `4.46m`.

amitu: Cant we do this using `fbt` tests? Maybe add a --ua (user agent), req.ua() will look at cli instead of user agent header in test mode.

harshdoesdev: when using `fbt` tests, `fastn serve` was giving `actix-web stack overflow` error.

amitu: Probably because we are doing debug build in fbt. Checkout `fbt.p1` file in ``fastn-core/tests` folder, it is using `--profile dev`. If you change `profile` to `release` our build time will go up, release build is slower than dev build, but one release build is faster than one release build + one dev build, so overall it would be faster than your current approach.

harshdoesdev: Cool! Got it.

amitu:
@harshdoesdev please resolve conflicts here.

PR #1751, opened 2024-02-05
Config read

amitu:


harshdoesdev:
@amitu, can we remove the config read in a different PR? We were planning to make some changes in `ft` for the HTML cache, and this is required for that.

PR #1752, opened 2024-02-05
Feat/environment

siddhantk232:
Use `fastn_ds::DocumentStore` to read environment variables instead of `std::env::var`. This is done for compatibility with our internal use of this framework.

I have also removed the unused `get_user_data_from_cookies`. We can always get it back from git history if we need to.


fastn-core/src/auth/routes.rs
```diff
@@ -11,24 +12,27 @@ pub async fn login(
     let provider = req.q("provider", "github".to_string())?;
 
     match provider.as_str() {
-        "github" => fastn_core::auth::github::login(req, next).await,
+        "github" => fastn_core::auth::github::login(ds, req, next).await,
         // client should handle redirects to next for email_password login
-        "emailpassword" => fastn_core::auth::email_password::login(req, db_pool, next).await,
+        "emailpassword" => fastn_core::auth::email_password::login(req, ds, db_pool, next).await,
```
amitu, line 17: Make this `email-password`.

amitu (approved):

PR #1754, opened 2024-02-06
cmd: Added sub-command `--check` in `fastn update`

harshdoesdev:
We are adding `--check` flag in the `fastn update` command. It will return with failure exit code of `.packages` is not in sync with `FASTN.ftd` and if it tries to perform a write operation during update check.

Discussion Link: [fastn update --check](https://github.com/orgs/fastn-stack/discussions/1740)

### DIFF
--- a/fastn-core/tests/20-fastn-update-check/cmd.p1
+++ b/fastn-core/tests/20-fastn-update-check/cmd.p1
@@ -1,9 +1,14 @@
 -- fbt:
 cmd: $FBT_CWD/../target/debug/fastn --test update --check
-exit-code: 1
+exit-code: 7
 
 -- stdout:
 
 -- stderr:
 
-FastnCoreError(UpdateError { message: "Check error: Write operation detected during check for package 'fastn-community.github.io/bling' at '<cwd>/.packages/fastn-community.github.io/bling/Changelog.md'" })
+Error: Out of Sync Package
+
+The package 'fastn-community.github.io/bling' is out of sync with the FASTN.ftd file.
+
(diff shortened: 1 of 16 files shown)


fastn-core/tests/20-fastn-update-check/cmd.p1
```diff
@@ -0,0 +1,9 @@
+-- fbt:
+cmd: $FBT_CWD/../target/debug/fastn --test update --check
+exit-code: 1
+
+-- stdout:
+
+-- stderr:
+
+FastnCoreError(UpdateError { message: "Check error: Write operation detected during check for package 'fastn-community.github.io/bling' at '<cwd>/.packages/fastn-community.github.io/bling/Changelog.md'" })
```
amitu, line 9: @harshdoesdev this is a very bad error message to show to end users.

fastn-core/tests/20-fastn-update-check/cmd.p1
```diff
@@ -0,0 +1,9 @@
+-- fbt:
+cmd: $FBT_CWD/../target/debug/fastn --test update --check
+exit-code: 1
```
amitu, line 3: Return Exit code 7 to mean check failure. Every other error also returns exit code 1, so end user will have to look at error message instead of exit code to see what happened.

fastn-core/tests/20-fastn-update-check/cmd.p1
harshdoesdev: I have updated the error message

amitu (approved):

PR #1757, opened 2024-02-08
Feat/auth updates

siddhantk232:
- make the onboarding step opt-in using environment variables. The env var `FASTN_AUTH_ADD_ONBOARDING_STEP` is introduced. The onboarding page is rendered only when this is set.
- Remove the `auth` feature flag.
- support both email/username for login flow. The `username` key can be either email or username.


amitu:
Hey @siddhantk232, do some more manual or automated testing. Yesterday I was having issues with signup flow, and github callback flow. In the error cases I got wrong messages, and sometimes username was taken for which I got error (or possibly no response). 

Also if you modify a field, the error message for that field should go away.

amitu (approved):

PR #1760, opened 2024-02-09
fix: remove manifest.json reads from fs fetch

harshdoesdev:
Removing `manifest.json` reads for dependency packages when resolving static file requests is the first step. The next step would be to update the logic for `ftd.image-src` so that if a particular variant of an image does not exist, it should not attempt to fetch it.


amitu:
@harshdoesdev is this ready?

harshdoesdev:
> @harshdoesdev is this ready?

Yes, this can be merged.

amitu (approved):

PR #1763, opened 2024-02-13
Add neovim syntax highlighting for ftd

xypnox:
Basic highlighting for neovim.

2024-02-13-132854_1281x1045_scrot


siddhantk232:
This is super cool :partying_face: 
Curious on the decision to use lua for this tho. You could have done this in vimscript which would enable `vim` users to also use this.

amitu:
> Curious on the decision to use lua for this tho. You could have done this in vimscript which would enable vim users to also use this.

We should do that, I suppose he either likes lua, or he got some other syntax thing, and just tweaked it. More PRs welcome :-)

PR #1768, opened 2024-02-14
Auth/reset password and some auth changes

siddhantk232:
This PR introduces a reset-password feature:

- Authenticated users can change their password by going to `/-/auth/set-password/`
- Unauthenticated (logged out) users can request password reset by goind to `/-/auth/forgot-password/`. This sends an email with a link to set their password.

## Auth changes

- `/-/auth/create-account/`: requires two new json body params, `password2` and `accept_terms`. See the [integration tests](https://github.com/fastn-stack/fastn/pull/1768/files#diff-b199dbb694cdbfc1f58506650f46a61d158e135705ee870c814e4f1773fc885d) to learn what has changes in request body.
- `/-/auth/create-account/`: redirect to `/-/auth/resend-confirmation-email/` on successful requests. This same route can be used to re-request confirmation emails in cases where user does not receive email.
- `FASTN_ENABLE_EMAIL`: new environment variable for use during development. This will be used to log emails to the console instead of actually sending it. This removes the need of `FASTN_SMTP*` environment variables.
- `req_config.request.ud()` checks on server session before returning the user. This enables invalidation of logged in users on password reset.

## Refactor

- [`Route`](https://github.com/fastn-stack/fastn/pull/1768/files#diff-423ef91702362f2b503ae4afdab5323f30cafb00e976d47c1b23b74aa2b70b3e) enum to keep track of auth related routes.

### What happens when a github user tried to reset their password?

The `/-/auth/set-password`, used when authenticated, converts their github oauth account to a normal email_password account. After doing this, they can use both the github login and the email_password login method.

Same goes for `/-/auth/forgot-password/`, entering their github username/email will send an email with a link to set their password. Check https://github.com/siddhantk232/fastn-test-multiauth to test things out.


amitu:
@siddhantk232 there are some conflicts, please resolve. We have switched to using `.starttls_relay()` (SMTP with StartTLS extension) as it seems neither AWS SES, nor postmark work with `.relay()` (SMTPS). Ideally we should make both protocol and port configurable.

siddhantk232:
There's a way to disable sending emails in dev mode (using `FASTN_ENABLE_EMAIL`) so we can live with `.starttls_relay()` for a short while. I'll later add more options for these in `mail.rs`

Heulitig (approved):

PR #1784, opened 2024-02-26
feat: kw-args

harshdoesdev:
Discussion: [`kw-args` on components](https://github.com/orgs/fastn-stack/discussions/1764)

Example:

```ftd
-- component foo:
string bar:
kw-args args:

-- ftd.text: $foo.bar

-- end: foo


-- foo:
bar: hello
baz: world
```


amitu:
@harshdoesdev in the example code always include some edge cases to prove the point, eg here include non kw-args to immediately answer everyons obvious answer if we can use both.

amitu:
Also how much work is it to implement this on record as well? Do not implement it if it is not trivial amount of work.

harshdoesdev:
> Also how much work is it to implement this on record as well? Do not implement it yet if it is not trivial amount of work.

It shouldn't be too much work. The logic should be similar to that of components. I'll try to implement it once I've implemented the kw-args for components.

Heulitig (approved):

PR #1785, opened 2024-02-26
feat(auth): option to disable strong password validation

siddhantk232:
Add option to disable strong password validation in `/-/auth/create-account/`.
This can be disabled using the `FASTN_DISABLE_STRONG_PASSWORD_CHECK` env var:

- `true`: disable validation
- `false` (or unset): use [`zxcvbn`](https://docs.rs/zxcvbn/2.2.2/zxcvbn/feedback/struct.Feedback.html) to check for weak passwords

Also removes some `dbg!` calls in `set_password.rs`.

### DIFF
--- a/integration-tests/_tests/04-multi-endpoint-test.test.ftd
+++ b/integration-tests/_tests/04-multi-endpoint-test.test.ftd
@@ -13,7 +13,7 @@ fastn.assert.eq(fastn.http_response["data"], "Hello, World!");
 ;; Mountpoint: /ftd/* -> Endpoint: http://fastn.com/ftd/*
 -- fastn.get: Fetching content from fastn.com
 url: /ftd/column/
-http-status: 308
+http-status: 301
 
 -- fastn.get: Redirect to google
 url: /goo/
--- a/fastn-core/src/auth/email_password/create_account.rs
+++ b/fastn-core/src/auth/email_password/create_account.rs
@@ -60,12 +60,24 @@ pub(crate) async fn create_account(
         )
     };
 
-    if let Err(e) = user_payload.validate_args((
+    if let Err(mut e) = user_payload.validate_args((
         user_payload.username.as_str(),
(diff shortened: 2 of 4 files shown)


integration-tests/_tests/04-multi-endpoint-test.test.ftd
```diff
@@ -13,7 +13,7 @@ fastn.assert.eq(fastn.http_response["data"], "Hello, World!");
 ;; Mountpoint: /ftd/* -> Endpoint: http://fastn.com/ftd/*
 -- fastn.get: Fetching content from fastn.com
 url: /ftd/column/
-http-status: 308
+http-status: 301
```
amitu, line 16: Why this change? 307/308 are better replacements of 301/302.

amitu:
This is because of Nandhini's comments? I am not yet sure if FifthTry should use this feature. I am okay with this being merged in general.

integration-tests/_tests/04-multi-endpoint-test.test.ftd
siddhantk232: I did this because the CI was failing but not because of the changes introduced in this PR. I couldn't track down the change that caused it

amitu: Got it.

siddhantk232:
> This is because of Nandhini's comments? I am not yet sure if FifthTry should use this feature. I am okay with this being merged in general.

Nope. I added this as a nice option to not have to come up with strong creative passwords when you're in a development environment. I remember you requesting this feature too

amitu:
> to not have to come up with strong creative passwords when you're in a development environment

Got it. So this is a development feature. We as a convention add either `DANGER_` (if it is a prod variable that you should not set normally) or `DEBUG_` (for development/debug variables) as prefix. FASTN_DISABLE_STRONG_PASSWORD_CHECK

PR #1788, opened 2024-02-27
feat: header in http processor

harshdoesdev:
Users can pass request headers as processor arguments. Header keys are keys that match this pattern: `$header-[key-name]$`. For example, `$header-authorization$` can be used to set the `Authorization` header of a request (as suggested by @Arpita-Jaiswal).

```ftd


-- create-post-response response:
$processor$: pr.http
method: POST
url: /blog/new/
$header-authorization$: Bearer auth-token


-- ftd.text: $response.message
```


amitu:
@harshdoesdev lets make this: `$header-authorization$: Bearer auth-token` instead. If tomorrow someone wants to send `header-name` or something as normal POST parameter they will end up sending wrong header.

fastn-core/src/http.rs
```diff
@@ -1016,3 +1016,14 @@ mod test {
         Ok(())
     }
 }
+
+static HEADER_KEY_PATTERN: once_cell::sync::Lazy<regex::Regex> =
+    once_cell::sync::Lazy::new(|| regex::Regex::new(r"\$header-(\w+)\$").unwrap());
```
amitu: Don't use regex for this, write a split_once based parser.

harshdoesdev: Okay

PR #1790, opened 2024-02-27
Logical logging

Arpita-Jaiswal:


fastn-ds/src/lib.rs
```diff
@@ -143,7 +124,14 @@ impl DocumentStore {
...
-    pub async fn capture_event(&self, _event: fastn_ds::Event) -> Result<(), EventError> {
+    pub async fn capture_event(
+        &self,
+        _okind: String,
+        _ekind: String,
+        _outcome: String,
+        _outcome_data: String,
+        _data: serde_json::Value,
+    ) -> Result<(), EventError> {
         Ok(())
```
amitu, line 135: Do a logging here.

fastn-ds/src/lib.rs
```diff
@@ -124,8 +124,8 @@ impl DocumentStore {
         }
     }
 
-    pub async fn capture_event(
-        &self,
+    pub fn capture_event(
+        &mut self,
```
amitu, line 128: Why do we need this? Can cause problem everywhere.

amitu: We should use Dashmap instead of Hashmap if you are worried about collecting data. It allows modifications on read only references.

PR #1792, opened 2024-02-29
minor fix: error on kw-arg values other than string

harshdoesdev:


amitu:
@harshdoesdev add test case for non string values:

```ftd
-- component foo:
integer kw-args data:

-- end: foo
```

Is this supported? If so write a test. Also write Rust test to extract the value of `data` from ftd document.

Heulitig (approved):

harshdoesdev:
> @harshdoesdev add test case for non string values:
> 
> ```
> -- component foo:
> integer kw-args data:
> 
> -- end: foo
> ```
> 
> Is this supported? If so write a test. Also write Rust test to extract the value of `data` from ftd document.

This syntax is not yet supported, only string values are supported.

amitu:
@harshdoesdev is there a test for extracting the data in Rust? If not create it.

harshdoesdev:
> @harshdoesdev is there a test for extracting the data in Rust? If not create it.

I have written an interpreter test for this: https://github.com/fastn-stack/fastn/pull/1793.

amitu:
@harshdoesdev I want the Rust code snippet someone has to use to extract data.

harshdoesdev:
> @harshdoesdev I want the Rust code snippet someone has to use to extract data.

Got it.

harshdoesdev:
> @harshdoesdev I want the Rust code snippet someone has to use to extract data.

@amitu I have added Rust tests for the `kw-args` feature and reading its values: https://github.com/fastn-stack/fastn/pull/1796

PR #1796, opened 2024-02-29
add: kwargs test and helper methods

harshdoesdev:
Adds Rust test for `kw-args` feature and a few helper methods for retrieving kwarg values.


ftd/src/interpreter/test.rs
```diff
@@ -258,3 +258,41 @@ fn evalexpr_test() {
...
+    // TODO: resolve value based on kwarg kind
+    let data: std::collections::BTreeMap<String, String> = kwargs
+        .iter()
+        .map(|(name, property_value)| {
+            (
+                name.to_string(),
+                property_value.to_value().get_string_data().unwrap(),
+            )
+        })
+        .collect();
```
amitu, line 294: Can you make this two lines?

harshdoesdev: the first few lines are boilerplate code. Do you mean the `let data = ..` line?

amitu: I mean can you take lines I have highlighted and make it two lines. Remove the boilerplace code.

ftd/src/interpreter/test.rs
```diff
@@ -258,3 +258,41 @@ fn evalexpr_test() {
...
+    -- ftd.text: Hello world
+
+    -- end: foo
+
+    -- foo:
+    bar: Hello
+    baz: World
+"#,
+    )
+    .unwrap_or_else(|e| panic!("{:?}", e));
```
amitu, line 278: This is also wrong. What is the code end user is supposed to write, write that code. Is interpret_helper in your opinion the method we should expose?

amitu: Point is not to test the "public API" for getting data. You have built a feature to extract data, just show me how am I supposed to use it.

ftd/src/interpreter/test.rs
```diff
@@ -258,3 +258,31 @@ fn evalexpr_test() {
...
+    -- foo:
+    bar: Hello
+    baz: World
+"#,
+    )
+    .unwrap();
+
+    let tdoc = doc.tdoc();
+    let instructions = doc.get_instructions("foo#foo");
+    let instruction = instructions.first().unwrap();
```
amitu, line 282: This is still not good. tdoc is temporary doc, and is internal property. Can you write a new crate, that uses ftd crate a dependency, and see what code you will write. Test against that code.

ftd/src/interpreter/test.rs
```diff
@@ -261,27 +191,26 @@ fn evalexpr_test() {
...
+        "#,
     )
     .unwrap();
 
-    let tdoc = doc.tdoc();
-    let instructions = doc.get_instructions("foo#foo");
-    let instruction = instructions.first().unwrap();
-
-    let data = instruction.get_kwargs("data", &tdoc).unwrap();
+    let component = doc.get_component_by_id("foo#fizz", "test").unwrap();
```
amitu, line 212: I would prefer if first argument was not needed. It should act like dom.get_element_by_id() in HTML (its playing exactly the same role, so why not copy their obviously superior behaviour).

amitu (approved):
Good job!

PR #1798, opened 2024-03-01
logical logging

Heulitig:
[Status](https://github.com/orgs/fastn-stack/discussions/1801)


fastn-core/src/http.rs
```diff
@@ -91,6 +91,45 @@ pub fn ok_with_content_type(
         .body(data)
 }
 
+#[derive(Debug, Default, Clone)]
+pub enum LogLevel {
+    #[default]
+    Info,
+    Warning,
+    Error,
+}
```
amitu, line 100: Nopes, lets be more logical. This is "generic logging". By logical we mean "precise logging" or whatever. Like we should have an exact enum of all log messages we can emit.

amitu: ekind-okind decide if something is info warning etc. Also error is captured as outcome, and error details have to go to outcome-data etc.

fastn-core/src/auth/email_password/set_password.rs
```diff
@@ -216,6 +217,7 @@ pub(crate) async fn forgot_password_request_success(
     req.log_with_no_site(
```
amitu, line 217: Remove the site business from here. This package will not know about site/org etc.

fastn-core/src/auth/email_password/login.rs
```diff
@@ -24,10 +25,37 @@ pub(crate) async fn login(
...
+        return match fastn_core::package::package_doc::read_ftd(
+            req_config, &main, "/", false, false,
+        )
+        .await
+        {
+            Ok(resp) => {
+                // [SUCCESS] logging
+                req.log(
+                    fastn_core::log::EventKind::Auth(fastn_core::log::AuthEvent::Login),
+                    fastn_core::log::EntityKind::Myself,
```
amitu, line 37: Since all auth related log will be done on myself, this Myself is useless. Remove it. 

Also I would prefer hard coding the event names, eg "login" instead of `fastn_core::log::EventKind::Auth(fastn_core::log::AuthEvent::Login)`, we will never be doing a match on this in Rust, this is *ONLY* going to be used in dashboard etc, and it would be currently a torture to find out what do put in where clause in my SQL query looking at this code.

fastn-core/src/auth/github/mod.rs
```diff
@@ -29,9 +30,22 @@ pub async fn login(
...
-        .await
-        .unwrap()
+    let github_client = match fastn_core::auth::github::utils::github_client(ds).await {
+        Ok(client) => client,
+        Err(e) => {
+            // [ERROR] logging (Environment Error)
+            let log_err_message = format!("environment: {:?}", &e);
+            req.log(
+                "github-login",
+                fastn_core::log::OutcomeKind::error_descriptive(log_err_message),
```
amitu, line 40: Here it is not clear what is the outcome. Error message is too never a good thing as it keeps changing, we need all this data for querying not for human consumption.

fastn-core/src/auth/github/mod.rs
```diff
@@ -73,44 +88,129 @@ pub async fn callback(
     // ask to update details by giving a form
     // redirect to next for now
     if fastn_core::auth::utils::is_authenticated(req) {
+        // [SUCCESS] logging: already authenticated
+        let log_success_message = "user: already authenticated".to_string();
```
amitu, line 92: Do log this kind of text, use text like `user-already-authenticated`.

amitu: Are you not setting the outcome? Outcome must be "user-error" in this case, and if I am interested in precisely what user error then I will want user-already-authenticated string. Normally I want to see how many user error happened today, vs how many system error.

amitu: Read the comments on Event class definition in models.py file.

PR #1800, opened 2024-03-04
feat(pr/request-data): add support default values

siddhantk232:
`request-data` works with optional string variables and string variables
with default values.

```ftd

;; if ?code=smth in url then $code is "smth"
;; else $code is NULL
-- optional string code: 
$processor$: pr.request-data

;; if ?code=smth in url then $code is "smth"
;; else $code is "default"
-- optional string code: default
$processor$: pr.request-data

;; if ?name=smth in url then $name is "smth"
;; else $name is "default"
-- string name: default
$processor$: pr.request-data

;; if ?err=smth in url then $err is "smth"
;; else an error will be thrown by fastn
-- string err:
$processor$: pr.request-data
```


amitu:
Write some tests also.

siddhantk232:
> Write some tests also.

could you link some existing tests that I can learn from?

siddhantk232:
Added some tests. Not sure if this is the correct way to test this tho, feel like I should've gone for integration tests.

PR #1802, opened 2024-03-08
feat: package lock and download-on-demand

harshdoesdev:
We are implementing package locking in fastn, which will allow us to keep packages in sync when they are deployed to production. Additionally, we are reintroducing download-on-demand to optimize the number of packages and assets that we download. We will perform the download-on-demand only in the `dev` mode.

Discussion: [Package Lock and Download-on-demand](https://github.com/FifthTry/business/discussions/327)

Todos:
- [x] `.packages -> .fastn/package-cache`
- [ ] Implement Download-on-demand in `--dev` mode
- [ ] Generate `.fastn/lock.json` lockfile
- [ ] Implement two-way merge for packages which usages contents of `.fastn/packages` and `.fastn/package-cache`


amitu:
@harshdoesdev rename the folder to `.fastn/package-cache`. Rationale, what would package-cache store if packages-cache stores multiple packages? Do you call it user table or users table? If users table contain multiple users, then you will say user table contains single user. Or user database, should it be called users database? How about book store? Is book store a place where you JUST SINGLE book and all book stores should be called books stores? How about movie theatre? Should it be movies theatre? Can repair shop, cars repair shop?

PR #1805, opened 2024-03-12
fastn_ds::http()

Heulitig:


fastn-ds/src/lib.rs
```diff
@@ -246,6 +267,86 @@ impl DocumentStore {
     pub async fn env(&self, key: &str) -> Result<String, EnvironmentError> {
         std::env::var(key).map_err(|_| EnvironmentError::NotSet(key.to_string()))
     }
+
+    // This method will connect client request to the out of the world
+    #[tracing::instrument(skip(req, extra_headers))]
+    pub async fn http<T>(
```
amitu, line 273: Don't think we should make this generic. Remove the generic parameter and make it concrete type.

amitu (changes requested):

PR #1809, opened 2024-03-13
moved http processor functions to fastn_ds

Heulitig:


fastn-ds/src/lib.rs
```diff
@@ -359,6 +364,141 @@ impl DocumentStore {
...
+            }
+
+            return Ok((Err(fastn_ds::DSError::GenericError { message }), cookies));
+        }
+        tracing::info!(msg = "returning success", url = url);
+        Ok((Ok(res.bytes().await?.into()), cookies))
+    }
+
+    #[tracing::instrument(skip_all)]
+    pub async fn http_post_with_cookie(
```
amitu, line 444: Can you make them delegate these functions to the already created http method so I just have to overwrite http and not many functions?

amitu (dismissed):

PR #1813, opened 2024-03-19
feat: add GET support in ftd.http

siddhantk232:
Add support for sending GET requests in `ftd.http`. Params are transformed into
query string or request body depending upon the type of request.

```ftd
-- ftd.http-options opts:
redirect: follow
method: GET
fastn-module: test-multiauth

-- void send-req():
ftd.http-options opts: $opts

ftd.http(
    "https://jsonplaceholder.typicode.com/comments",
    opts,
    ("postId", 1)
)
```

the above snippet will send request to
`https://jsonplaceholder.typicode.com/comments?postId=1` with the given
options.

```ftd
-- ftd.http-options opts:
redirect: follow
method: POST ;; ONLY THIS IS CHANGED FROM THE ABOVE SNIPPET
fastn-module: test-multiauth

-- void send-req():
ftd.http-options opts: $opts

ftd.http(
    "https://jsonplaceholder.typicode.com/comments",
    opts,
    ("postId", 1)
)
```

the above snippet will send request to
`https://jsonplaceholder.typicode.com/comments` with request body as
`{"postId": 1}`.

- existing query params are preserved.
- if the url already has a query param with the same name, it will be overridden.


fastn-js/js/ftd.js
```diff
@@ -244,7 +247,14 @@ const ftd = (function () {
...
         }
 
-        init.body = JSON.stringify(request_json);
+        if (method === "POST") {
+            init.body = JSON.stringify(request_json);
+        }
+
+        if (method === "GET") {
+            const params = new URLSearchParams(request_json);
+            url += `?${params.toString()}`;
```
amitu, line 256: This wont work if URL already contains some query params. Ideally we should preserve them and add these params.

siddhantk232: good catch, thanks!

siddhantk232: fixed in https://github.com/fastn-stack/fastn/pull/1813/commits/c86a7c7cc5dc7dfb59c08e6c153bfe34a52856eb

amitu (approved):

PR #1814, opened 2024-03-20
some code cleanups

amitu:


amitu:
I have verified that fastn binary built using this branch is able to serve ui.fifthtry.com repo.

PR #1823, opened 2024-03-26
use rquickjs for release builds

siddhantk232:
This fixes the release pipeline. I'll later add a PR that'll add a dynamic binary that uses deno_core in the list of release artifact so that we can use that in our install script.

- to build with deno_core use the `deno-core` feature flag


harshdoesdev (dismissed):

amitu:
@siddhantk232 tests are failing.

amitu (dismissed):

PR #1824, opened 2024-03-26
toc processor body fix

Heulitig:


amitu:
@Heulitig write a failing test case which gets fixed because of this (else we will regress and keep fixing things like this).

PR #1829, opened 2024-04-02
feat/rm auth

siddhantk232:
# remove auth and related features

Authentication related feature and routes are removed, specifically:

- auth handlers for email_password and github OAuth providers which were
  working
- other untested OAuth providers (discord, twitter, telegram)
- Below listed processor yield an error when used, these used to provide "mixed
  privacy" features which was never full implemented nor documented anywhere.
  We will bring a revised version "mixed privacy" feature in future.
    - `document-readers`
    - `user-groups`
    - `user-group-by-id`
    - `get-identities`
    - `is-reader`

- **http::Request::ud is marked todo!()**: This is a major breaking change as
  we do use it in our internal codebase. There are three ways I can think of by
  which we can keep this function in fastn-core:

    1. we use fastn-community/auth as a dependency and use a function exposed by its rust
    code in here
    2. we use tokio-postgres and assume `fastn_user` table exists when FASTN_ENABLE_AUTH is
    set. ud() will also only work when FASTN_ENABLE_AUTH is set
    3. we suggest everyone to use fastn-community/auth as a dependency and use
       this function from there.
  
  I think we should either go with first or third option. Going with third
  means that fastn does not know or care about auth at all which imo is a good
  thing.

- remove diesel dep


amitu (approved):

amitu:
I agree with the third option. Let's try to use auth stuff at FifthTry the way we want our end customers to use it.

PR #1832, opened 2024-04-05
migration changes

Heulitig:


fastn-core/src/lib.rs
```diff
@@ -48,7 +48,7 @@ pub(crate) use tracker::Track;
 pub(crate) use translation::{TranslatedDocument, TranslationData};
 pub use {doc::resolve_foreign_variable2, doc::resolve_import};
 
-pub const FASTN_UI_INTERFACE: &str = "fastn-stack.github.io/fastn-ui";
+pub const FASTN_UI_INTERFACE: &str = "fastn-ui.fifthtry-site";
```
amitu, line 51: Why is it not .site? Also can we remove fastn ui interface altogether, we are not using it, when we start using it we can revive this.

PR #1843, opened 2024-04-26
record closure fixes

Heulitig:


amitu:
@Heulitig add a failing test case also (that fails on current main but passes after changes in this PR) so we do not have the same issue in future.

PR #1844, opened 2024-04-26
Fix: Using sqlite

Arpita-Jaiswal:


Cargo.toml
```diff
@@ -110,7 +110,7 @@ ftd-p1 = { path = "ftd-p1" }
 ftd-tc = { path = "ftd-tc" }
 ftd-ast = { path = "ftd-ast" }
 fastn-js = { path = "fastn-js" }
-ft-sys-shared = "0.1"
+ft-sys-shared = { path = "../../fifthtry/ft-sdk/ft-sys-shared", version = "0.1" }
```
amitu, line 113: I think we will have to publish a new version of ft-sys-shared as otherwise it won't work on github action.

amitu: I have published the alpha version.

PR #1853, opened 2024-05-03
wasi attempt

amitu:


amitu:
We are giving up on this approach tho. We were thinking of adding wasi not because we want to support wasi interface, but so that we can use https://docs.rs/quickjs-wasm-rs/latest/quickjs_wasm_rs/. On a second thought, running quickjs inside wasm is only a security measure, and that too dubious, and slows things down and complicates our code base. Till we have better idea, we are going to use quickjs directly.

PR #1870, opened 2024-05-19
Added ftd.string-field-with-default

Arpita-Jaiswal:


amitu:
@Arpita-Jaiswal the test is failing. Also can we zero out the JS from these? Like detect if we are running in test mode and not include the standard JS files, instead put a `__common__.js` or some such tag.

PR #1875, opened 2024-05-21
Fix: http processor url to full url

Arpita-Jaiswal:


amitu:
@Arpita-Jaiswal tests are failing.

PR #1894, opened 2024-06-04
fastn migration in progress

Arpita-Jaiswal:
- [x] https://github.com/FifthTry/ft-sdk/pull/43 
- [x] https://github.com/FifthTry/dotcom/pull/38
- [x] add `fastn_user` related migraitons
- [x] latest migration should be written on top of the list

### DIFF
--- a/fastn-core/src/commands/serve.rs
+++ b/fastn-core/src/commands/serve.rs
@@ -181,46 +181,50 @@ pub async fn serve(
     config: &fastn_core::Config,
     req: fastn_core::http::Request,
     only_js: bool,
-) -> fastn_core::Result<fastn_core::http::Response> {
+) -> fastn_core::Result<(fastn_core::http::Response, bool)> {
     if req.path() == "/-/auth/logout/" {
-        return Ok(clear_sid2(&req));
+        return Ok((clear_sid2(&req), false));
     }
 
     if let Some(endpoint_response) = handle_endpoints(config, &req).await {
-        return endpoint_response;
+        return endpoint_response.map(|r| (r, false));
     }
 
     if let Some(app_response) = handle_apps(config, &req).await {
-        return app_response;
+        return app_response.map(|r| (r, false));
     }
 
     if let Some(default_response) = handle_default_route(&req, config.package.name.as_str()) {
-        return default_response;
+        return default_response.map(|r| (r, true));
     }
 
     let path: camino::Utf8PathBuf = req.path().replacen('/', "", 1).parse()?;
 
     if let Some(r) = handle_redirect(config, &path) {
-        return Ok(r);
+        return Ok((r, false));
     }
 
     if fastn_core::utils::is_static_path(req.path()) {
-        return handle_static_route(req.path(), config.package.name.as_str(), &config.ds).await;
+        return handle_static_route(req.path(), config.package.name.as_str(), &config.ds)
+            .await
+            .map(|r| (r, true));
     }
 
-    let req_config = fastn_core::RequestConfig::new(config, &req, "", "/");
+    let mut req_config = fastn_core::RequestConfig::new(config, &req, "", "/");
 
-    serve_helper(req_config, only_js, path).await
+    serve_helper(&mut req_config, only_js, path)
+        .await
+        .map(|r| (r, req_config.response_is_cacheable))
 }
 
 #[tracing::instrument(skip_all)]
 pub async fn serve_helper(
-    mut req_config: fastn_core::RequestConfig,
+    req_config: &mut fastn_core::RequestConfig,
     only_js: bool,
     path: camino::Utf8PathBuf,
 ) -> fastn_core::Result<fastn_core::http::Response> {
     let mut resp = if req_config.request.path() == "/" {
-        serve_file(&mut req_config, &path.join("/"), only_js).await
+        serve_file(req_config, &path.join("/"), only_js).await
     } else {
         // url is present in config or not
         // If not present than proxy pass it
@@ -269,7 +273,7 @@ pub async fn serve_helper(
             }
         }
 
-        let file_response = serve_file(&mut req_config, path.as_path(), only_js).await;
+        let file_response = serve_file(req_config, path.as_path(), only_js).await;
 
         tracing::info!(
             "before executing proxy: file-status: {}, path: {}",
@@ -284,7 +288,7 @@ pub async fn serve_helper(
         return shared_to_http(r);
     }
 
-    for cookie in req_config.processor_set_cookies {
+    for cookie in &req_config.processor_set_cookies {
         resp.headers_mut().append(
             actix_web::http::header::SET_COOKIE,
             actix_web::http::header::HeaderValue::from_str(cookie.as_str()).unwrap(),
@@ -576,7 +580,7 @@ async fn actual_route(
     tracing::info!(method = req.method().as_str(), uri = req.path());
     let req = fastn_core::http::Request::from_actix(req, body);
 
-    serve(config, req, false).await
+    serve(config, req, false).await.map(|(r, _)| r)
 }
 
 #[tracing::instrument(skip_all)]
--- a/fastn-core/src/commands/serve.rs
+++ b/fastn-core/src/commands/serve.rs
@@ -186,7 +186,7 @@ pub async fn serve(
         return Ok(clear_sid2(&req));
     }
 
-    handle_migrations(config, &req).await?;
+    handle_migrations(config).await?;
 
     if let Some(endpoint_response) = handle_endpoints(config, &req).await {
         return endpoint_response;
@@ -491,13 +491,8 @@ async fn handle_static_route(
     }
 }
 
-async fn handle_migrations(
-    config: &fastn_core::Config,
-    req: &fastn_core::http::Request,
-) -> fastn_core::Result<()> {
-    let mut req_config = fastn_core::RequestConfig::new(config, &req, "", "/");
-
-    Ok(fastn_core::migrations::migrate(&mut req_config).await?)
+async fn handle_migrations(config: &fastn_core::Config) -> fastn_core::Result<()> {
+    Ok(fastn_core::migrations::migrate(config).await?)
 }
 
 async fn handle_endpoints(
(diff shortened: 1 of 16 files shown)


fastn-core/src/commands/serve.rs
```diff
@@ -186,6 +186,10 @@ pub async fn serve(
         return Ok(clear_sid2(&req));
     }
 
+    if let Some(endpoint_response) = handle_migrations(config, &req).await {
+        return endpoint_response;
```
amitu, line 190: Why would migration return a reponse? Also migration can error, so its return type must be `Result`.

fastn-core/src/commands/serve.rs
```diff
@@ -489,6 +493,26 @@ async fn handle_static_route(
...
+
+    let db = fastn_core::migrations::get_db_url(config).await;
+    fastn_core::migrations::create_migration_table(config, db.as_str()).await?;
+
+    fastn_core::migrations::migrate(&mut req_config, db.as_str()).await?;
+
+    Some(Ok(actix_web::HttpResponse::build(actix_web::http::StatusCode::OK)
+        .append_header(("Content-Type", "application/json"))
+        .body(serde_json::json!({"status": "success"}).to_string())))
+}
```
amitu, line 514: Nopes. Before we process any page we will run migration (this is the best we can do, ideally we should only run migration if migration section has changed, or we load the site for the first time etc, not sure how can we do it in fastn, maybe we can do this kind of optimisation in hostn).

fastn-core/src/library2022/processor/sql.rs
```diff
@@ -10,15 +10,7 @@ pub async fn process(
...
-        None => match config.config.ds.env("FASTN_DB_URL").await {
-            Ok(db_url) => db_url,
-            Err(_) => config
-                .config
-                .ds
-                .env("DATABASE_URL")
-                .await
-                .unwrap_or_else(|_| "fastn.sqlite".to_string()),
-        },
+        None => fastn_core::migrations::get_db_url(&config.config),
```
amitu, line 13: This should be a method on config, `config.get_db_url()`.

fastn-core/src/migrations.rs
```diff
@@ -0,0 +1,224 @@
...
+}
+
+async fn get_migrations_from_ftd(
+    req_config: &mut fastn_core::RequestConfig,
+) -> Vec<MigrationDataFTD> {
+    let migration_ftd = req_config
+        .config
+        .package
+        .fs_fetch_by_file_name("migration.ftd", None, &req_config.config.ds)
+        .await?;
```
amitu, line 85: Nopes, migraiton must be `FASTN.ftd` for now. If we allow this we have to open and parse one more file on every page view.

fastn-core/src/migrations.rs
```diff
@@ -0,0 +1,224 @@
...
+            }
+            None => {
+                migrations_to_apply.push(m_ftd.clone());
+            }
+        }
+    }
+    migrations_to_apply
+}
+
+async fn get_migration_from_sql(
```
amitu, line 167: When you take a flight from hometown, do you come FROM flight or come from hometown? Do you come by flight or come by hometown? Do you come from hometown by flight or come by hometown from flight?

fastn-core/src/migrations.rs
```diff
@@ -0,0 +1,224 @@
...
+                .get(1)
+                .expect("fastn_migration::migration_name not found.")
+                .clone(),
+        )
+        .unwrap();
+        migration_details.push(MigrationDataSQL {
+            number: migration_number,
+            name: migration_name,
+        });
+    }
```
amitu, line 211: This is to check if migraiton names etc have not changed. Let’s not do this, as this is too much work on every page load. Find the latest migration number, and compare it with number of migrations in fastn.ftd file, and if they differ, and it comes time to apply an actual migration, we first do rigerous check and then apply migraiton.

fastn-core/src/migrations.rs
```diff
@@ -0,0 +1,224 @@
...
+
+struct MigrationDataSQL {
+    number: i64,
+    name: String,
+}
+
+async fn find_migrations_to_apply(
+    migration_ftd: &[MigrationDataFTD],
+    migration_sql: &[MigrationDataSQL],
+) -> Vec<MigrationDataFTD> {
```
amitu, line 142: This function should not be async.

fastn-core/src/commands/serve.rs
Arpita-Jaiswal: Earlier we were planning to run migration when the [user visits /migration/ url](https://github.com/orgs/fastn-stack/discussions/1881)

amitu: Storing it in FASTN.ftd is better design as otherwise people will have to explicitly go to that URL to triggger migraiton.

Arpita-Jaiswal: @amitu But this saves us from running migrations repeatedly and helps us do more expensive checks because this won't be happening so frequently.

fastn-core/src/library2022/processor/sql.rs
Arpita-Jaiswal: Done

fastn-core/src/commands/serve.rs
```diff
@@ -489,6 +491,10 @@ async fn handle_static_route(
     }
 }
 
+async fn handle_migrations(config: &fastn_core::Config) -> fastn_core::Result<()> {
+    Ok(fastn_core::migrations::migrate(config).await?)
+}
```
amitu, line 496: Not a useful function, call `fastn_core::migrations::migrate(config)` directly from the `serve` function.

fastn-core/src/migrations.rs
```diff
@@ -0,0 +1,210 @@
...
+
+CREATE TABLE IF NOT EXISTS
+    fastn_migration
+(
+    id               INTEGER PRIMARY KEY,
+    app_name         TEXT NOT NULL,
+    migration_number INTEGER NOT NULL UNIQUE,
+    migration_name   TEXT NOT NULL,
+    applied_on       INTEGER NOT NULL
+) STRICT;
```
amitu, line 11: Create another PR in ft-sdk and remove the corresponding table from there, and link to this PR. We should merge both PR together.

fastn-core/src/migrations.rs
```diff
@@ -0,0 +1,210 @@
...
+
+"#;
+
+pub(crate) async fn migrate(config: &fastn_core::Config) -> Result<(), MigrationError> {
+    let db = config.get_db_url().await;
+    create_migration_table(config, db.as_str()).await?;
```
amitu, line 17: In general if you are passing `x`, and a data derived from `x`, the consider just passing `x`.

amitu: Make `migrate()` a method of `fastn_core::Config`. Same as `has_new_migration`.

fastn-core/src/migrations.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    if !has_new_migration(config, db.as_str()).await? {
```
amitu, line 19: Rename to `has_unapplied_migrations()`.

fastn-core/src/migrations.rs
amitu: Leave the implementations in this file only. 

```rs
impl Config {
    pub(crate) fn migrate() {}
    pub(crate) fn has_unapplied_migrations() {}
}
```

fastn-core/src/migrations.rs
```diff
@@ -0,0 +1,210 @@
...
+                "#,
+                config.package.name
+            )
+            .as_str(),
+            vec![],
+        )
+        .await?;
+
+    match results.len() {
+        0 => Ok(true),
```
amitu, line 84: Overall feature may work fine, but this statement is wrong. If there are no migrations in the db it does not on its own mean we have unapplied migrations. 

Also in general if some task is done once in one way, unless we have strong reason, we should do it the next time in same way. The implementation of migration in ft-sdk is differnt from there. So unless you have found fault in that implementation (eg if you feel this implementation is technically better) I prefer we use the same Algo as there. Same function names etc, same behaviour etc.

fastn-core/src/commands/serve.rs
```diff
@@ -186,6 +186,8 @@ pub async fn serve(
         return Ok(clear_sid2(&req));
     }
 
+    handle_migrations(config).await?;
```
amitu, line 189: On a second thought, lets not call this function here, we can apply migration when Config object is first constructed. Since we re-use config across requests this will be more optimal.

fastn-core/src/migrations.rs
Arpita-Jaiswal: @amitu I think it should be in `ft-sys-shared`.

fastn-core/src/commands/serve.rs
Arpita-Jaiswal: done

fastn-core/src/migrations.rs
Arpita-Jaiswal: done

fastn-core/src/commands/serve.rs
Arpita-Jaiswal: done

fastn-core/src/migrations.rs
Arpita-Jaiswal: Not following this in favor of https://github.com/fastn-stack/fastn/pull/1894#discussion_r1627130340

fastn-core/src/migrations.rs
Arpita-Jaiswal: Not following this in favor of https://github.com/fastn-stack/fastn/pull/1894#discussion_r1627130340

fastn-core/src/migrations/mod.rs
```diff
@@ -112,11 +138,10 @@ async fn find_latest_applied_migration_number(
                     FROM
                         fastn_migration
                     WHERE
-                        app_name = '{}'
+                        app_name = '{app_name}'
```
amitu, line 141: Nopes, use proper argument here.

amitu (approved):

PR #1896, opened 2024-06-06
inlining fbt

amitu:


amitu:
@Arpita-Jaiswal lets include fbt in this repo as otherwise we compile ftd twice. The crate is small enough (and does not change much so) that we can maintain from here.

PR #1903, opened 2024-06-11
Added code description for http processor to handle response code

Arpita-Jaiswal:


fastn-core/src/library2022/processor/http.rs
```diff
@@ -142,6 +142,8 @@ pub async fn process(
                     Ok(200) => (),
                     Ok(code) => {
                         req_config.processor_set_response = Some(r);
+                        // This error is to short-circuit it and be handled by
+                        // `fastn_core::commands::serve::shared_to_http`
```
amitu, line 146:
```suggestion
                        // We return an error here, this error will not be shown to user,  but is created to to 
                        // short-circuit further processing. the `.processor_set_response` will be handled by
                        // `fastn_core::commands::serve::shared_to_http`
```

PR #1924, opened 2024-07-09
correctly render `srcdoc` attribute for `ftd.iframe`

siddhantk232:
Providing a caption ([`srcdoc`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement/srcdoc)) to `ftd.iframe` renders a `src` attribute. This is fixed in this PR.

The expected output html for the following ftd code:

```ftd
-- ftd.iframe

<h1>hello world</h1>
```

should be:

```
<iframe srcdoc="<h1>hello world</h1>" />
```

but it is (notice the attribute is `src`):

```
<iframe src="<h1>hello world</h1>" />
```

The will result in an invalid url error in browsers.


siddhantk232:
seems like the only test failed is `'tests::fbt' panicked at fastn-core/src/lib.rs:207:13`

PR #1936, opened 2024-07-29
Added session_id method in Request

Arpita-Jaiswal:
Earlier the `session_id` method in `RequestConfig` only read session from request header `X-FASTN-ACTOR`. Fix this by reading from cookie too. Also moved the logic to `Request::session_id` method, so that it can be used by `hostn::route` too.

Related PR: https://github.com/FifthTry/ft/pull/182

### DIFF
--- a/fastn-core/src/library2022/processor/user_details.rs
+++ b/fastn-core/src/library2022/processor/user_details.rs
@@ -8,7 +8,9 @@ pub async fn process(
     match ud(
         &req_config.config.ds,
         req_config.config.get_db_url().await.as_str(),
-        req_config.request.cookie(fastn_core::http::SESSION_COOKIE_NAME),
+        req_config
+            .request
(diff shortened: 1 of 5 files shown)


fastn-core/src/library2022/processor/user_details.rs
```diff
@@ -5,12 +5,16 @@ pub async fn process(
...
-    )
-    .await
+    match req_config
+        .config
+        .ds
+        .ud(
+            req_config.config.get_db_url().await.as_str(),
+            &req_config
+                .request
+                .cookie(fastn_core::http::SESSION_COOKIE_NAME),
```
amitu, line 15: This should also use the `session_id()` method.

Arpita-Jaiswal: Done. Thanks.

PR #1937, opened 2024-08-02
Added support for `diff` lang in `ftd.code`

Arpita-Jaiswal:
- Added `prism-diff.js`
- Done changes in `css` theme to support tags by `diff`


amitu:
@Arpita-Jaiswal the PR title is wrong.

PR #1945, opened 2024-10-03
sql-query: Return zero rows for IDE preview requests

siddhantk232:
Preview requests can come from other users and anonymous users. Running
sql-query for them is a security threat.


fastn-core/src/library2022/processor/sql.rs
```diff
@@ -6,11 +6,20 @@ pub async fn process(
     doc: &ftd::interpreter::TDoc<'_>,
     config: &mut fastn_core::RequestConfig,
     q_kind: &str,
+    preview_session_id: &Option<String>,
```
amitu, line 9: Change the signature to accept a Boolean, which indicates if query should be executed so in future when you do session id check, you do not have to bring that logic to this level.

amitu (changes requested):

fastn-core/src/library2022/processor/sql.rs
siddhantk232: done in https://github.com/fastn-stack/fastn/pull/1945/commits/dc187a525677361928fb7626a0d41f1e46a93608

fastn-core/src/library2022/mod.rs
```diff
@@ -304,9 +304,42 @@ impl Library2022 {
...
+            "sql-batch" => {
+                processor::sql::process(
+                    value,
+                    kind,
+                    doc,
+                    self,
+                    "sql-batch",
+                    preview_session_id.is_none(),
+                )
+                .await
```
amitu, line 341: Can't we put the complete logic here: 
```
if preview_session_id.is_none() { Default::default() } else { 
processor::sql::process(
                    value,
                    kind,
                    doc,
                    self,
                    "sql-batch",
                )
                .await
```

siddhantk232: Tried doing something similar in https://github.com/fastn-stack/fastn/pull/1945/commits/eb6dc4b6da57e70f3bfd48b3b099a671e106b421

amitu:
@siddhantk232 hopes. This will make the page fail to load as processor would not be found (not 100% sure, correct me if I am wrong). We want to show regular ftd page with zero results, page should render.

siddhantk232:
> @siddhantk232 nopes. This will make the page fail to load as processor would not be found (not 100% sure, correct me if I am wrong). We want to show regular ftd page with zero results, page should render.

Don't think so

(original website)
image

(website in preview)
image

amitu (approved):

PR #1947, opened 2024-10-03
Prepare for release version 0.4.76

siddhantk232:


amitu:
Do a cargo update as well. Also check if the rust version in toolchain is latest.

siddhantk232:
> Do a cargo update as well. Also check if the rust version in toolchain is latest.

Updated cargo deps. We're already using the latest rust (1.81.0).

amitu:
@siddhantk232 after you do cargo update in this repo, when you use this version in ft, do a cargo update there as well.

PR #1952, opened 2024-10-07
Remove markdown parsing from frontend

siddhantk232:
[marked](https://github.com/markedjs/marked/) was used to parse markdown
for each text node in the browser which was making things very slow.

This commit moves markdown parsing to the backend using the `comrak`
crate.

Some cleanups may be required that removes the marked.js files
completely from the codebase but it'll happen later once I get current
changes reviewed.

## `ftd` code that was used to test this

```ftd
-- ftd.text: 

Hello world `hellow` [link](https://google.com/)

\```rust
fn hello () -> {
    

    println!("works?");
    println!("works?");
    println!("works?");
    println!("works?");
    println!("works?");
    println!("works?");



}
\```

-- ftd.code:
lang: rust

#[derive(Debug)]
pub struct SetProperty {
    pub kind: PropertyKind,
    pub value: SetPropertyValue,
    pub element_name: String,
    pub inherited: String,
}

#[derive(Debug)]
pub enum SetPropertyValue {
    Reference(String),
    Value(fastn_js::Value),
    Formula(fastn_js::Formula),
    Clone(String),
}

impl fastn_js::SetPropertyValue {
    pub fn to_js(&self) -> String {
        self.to_js_with_element_name(&None)
    }
}
```

I couldn't see any visual changes before and after this commit. Attaching screenshots anyway:

(before this commit)
image

(after this commit)
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1575,9 +1575,11 @@ dependencies = [
 name = "fastn-js"
 version = "0.1.0"
 dependencies = [
+ "comrak",
  "fastn-grammar",
  "indoc",
  "itertools 0.13.0",
+ "once_cell",
  "prettify-js",
  "pretty",
  "quick-js",
--- a/fastn-core/src/utils.rs
+++ b/fastn-core/src/utils.rs
@@ -755,11 +755,9 @@ pub async fn replace_markers_2023(
             r#"
                 <script src="{}"></script>
                 <script src="{}"></script>
-                <script src="{}"></script>
                 <link rel="stylesheet" href="{}">
                 {}
             "#,
-            hashed_markdown_js(),
             hashed_prism_js(),
             hashed_default_ftd_js(config.package.name.as_str()),
             hashed_prism_css(),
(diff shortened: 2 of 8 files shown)


amitu:
@siddhantk232 how have you tested it so far? Is fastn.com, fifthtry, ide etc working with it?

amitu:
Can you also create some stress test / benchmark files, eg with 10K ftd.text nodes (with slightly different content; generated using Python) and do a before and after (backend and frontend timing) to see if it indeed made any improvements or is it a performance regression?

siddhantk232:
> @siddhantk232 how have you tested it so far? Is fastn.com, fifthtry, ide etc working with it?

I tested it with the ftd snippet mentioned in the [PR description](https://github.com/fastn-stack/fastn/pull/1952#issue-2570368690) which worked fine. fastn.com does have some visual differences (ftd code is not highlighted and other code is not indented properly with this change). I don't see any noticeable performance gains either:

[fastn 0.4.76 lighthouse report on `fastn.com/web-component/` ran locally with `fastn --trace serve --offline`](https://github.com/user-attachments/files/17291540/fastn_current.pdf)
[Same webpage but with current build of fastn that includes this change](https://github.com/user-attachments/files/17291542/report_fastn_dev.pdf)

@amitu

siddhantk232:
I don't think moving the code highlight part (prism.js) to the backend would make this any better. We should probably not go ahead with this for now.

Issue #552, opened 2023-01-25
Performance issue FTD 0.3 - kills server activity


gsalunke:
While working on a below file I observed stopped and below message seen:

```
[2023-01-25T04:46:55Z INFO ] "GET /builder/issues HTTP/1.1" 0.029471s 200 55892 127.0.0.1 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/109.0.0.0 Safari/537.36"
zsh: killed     fpm serve
```
<details>
<summary>FTD Code:</summary>

```ftd

-- card-container:

-- issue: gap: 32px seen on H1, H2, H3 headings
reported-on: 12 Jan 2023
github-link: https://github.com/ftd-lang/ftd/blob/main/Cheatsheet.md

-- issue.error-msg: 

`fpm.error no error message`

`fpm.error no error message`

`fpm.error no error message`


-- end: card-container










-- component issue:
caption issue:
optional string reported-on:
optional string error-msg:
optional string github-link:

-- card:

-- ftd.column:
width: fill-container

-- card-header:

-- issue-body:
issue: $issue.issue
reported-on: $issue.reported-on
error-msg: $issue.error-msg
github-link: $issue.github-link

-- end: ftd.column

-- end: card

-- end: issue


-- component card:
children card-wrap:

-- ftd.column:
width: fill-container
background.solid: #000000
spacing.px: 16
padding-vertical.px: 24
padding-horizontal.px: 24
border-bottom-width.px: 1
border-color: $inherited.colors.border-strong
margin-bottom.px: 24
children: $card.card-wrap

-- end: ftd.column

-- end: card










-- component card-container:
children card-child:

-- ftd.column:
width: fill-container
background.solid: #333333

-- ftd.column:
width.fixed.px: 1160
align-self: center
children: $card-container.card-child

-- end: ftd.column

-- end: ftd.column

-- end: card-container

;;card-toolkit








-- component card-header:

-- ftd.column:
width: fill-container

-- ftd.row:
width: fill-container
spacing.px: 28

-- ftd.text: Issue Description
role: $inherited.types.heading-small
color: $inherited.colors.text

-- ftd.text: Reported on
role: $inherited.types.heading-small
color: $inherited.colors.text
white-space: nowrap

-- ftd.text: Error Message
role: $inherited.types.heading-small
color: $inherited.colors.text

-- ftd.text: GitHub Code Link
[...]

PR #1975, opened 2024-10-15
Fix infinite loop stuck

Arpita-Jaiswal:
This PR introduces the `ftd::interpreter::InterpreterState::detect_cycle` function to detect cycles in the component processing sequence and raises an error if a cycle is found. It prevents infinite loops caused by cyclic dependencies during component processing. Components may be processed multiple times if they have unprocessed dependencies or if it's stuck in an infinite loop. However, this function ensures that no component is stuck in a recursive cycle by distinguishing between legitimate reprocessing (due to unresolved dependencies) and infinite loops (due to cyclic dependencies).

## Theory of cycle detection

A cycle is detected when a repeating pattern of components appears in the `in-process` list, for example:

```
 a -> b -> c -> d -> e -> c -> d -> e -> c -> d -> e
           ^              ^              ^
```

In this case, `c -> d -> e` is the repeating pattern, indicating a cyclic dependency that will never resolve.

##  How it works:
1. **Remove Already Processed Components**: The function begins by clearing any components that have already been fully processed from the `in-process` list to avoid redundant checks.

2. **Skip Already Processed Nodes**: It skips AST nodes that are either invocations, imports, or have already been processed (tracked via `self.bag`). This prevents unnecessary re-processing.

3. **Handling Multiple In-Process Entries**: If a component appears multiple times to be processed, it could either be due to unprocessed dependencies or a cyclic dependency (infinite loop). To handle this, we use `number_of_scan`. Every time a component is re-processed because of unprocessed dependencies, the `number_of_scan` is incremented. This allows us to differentiate between normal re-processing and an infinite loop.

4. **Cycle Detection**: The function then checks for cycles by comparing the current `in-process` sequence with its prior state:
    - It compares the last half of the `in-process` list to the preceding half to check for repeating 
      patterns (which would indicate a cycle).
    - If a cycle is detected, an error is raised by calling `self.construct_cycle_error(i)`.

5. **Completion**: If no cycle is detected, the function proceeds normally, and the component is processed without error.


amitu:
@Arpita-Jaiswal tests are failing

Arpita-Jaiswal:
> @Arpita-Jaiswal tests are failing

@amitu The PR is not complete yet.

amitu:
@Arpita-Jaiswal can you also take a look at this one: https://github.com/fastn-stack/fastn/issues/552

Also I think there was another issue related to partial renaming of component, eg component has been renamed, but old component name is still referenced in the body of component, causing infinite loop.

ftd/t/interpreter/32-recursion.ftd
```diff
@@ -0,0 +1,35 @@
...
+-- ftd.column:
+
+-- ftd.text: $print-tree.t.name
+
+-- print-tree: $obj
+for: $obj in $print-tree.t.children
+
+-- end: ftd.column
+
+-- end: print-tree
```
amitu, line 35: All these red signs, get rid of them. Add new line at the end of each file that doesn't have them.

amitu (approved):

amitu:
@Arpita-Jaiswal do a version bump, merge this PR and create a release.

PR #1976, opened 2024-10-15
Fix: Check for self referencing kinds in `record`, `function` and, `component`

siddhantk232:
```ftd
-- work:
smth: smth-idk

-- component work:
work smth:

-- ftd.text: something here

-- end: work
```

Line number 6 on the above code snippet would make `fastn` go in an
infinite loop. We now throw a `ParseError` for this that we can't parse
self referencing kinds.


siddhantk232:
Closing this as it's a duplicate of #1975. I wasn't sure of this change anyway

PR #2000, opened 2024-10-28
`fastn 0.5`

Arpita-Jaiswal:


v0.5/fastn-p1/src/parser/scanner.rs
```diff
@@ -134,18 +134,30 @@ impl Scanner {
 
     // returns the span from current position to the end of token
     pub fn token(&mut self, t: &'static str) -> Option<fastn_p1::Span> {
-        let mut count = 0;
-        for char in t.chars() {
+        // Get the length of the token to match
+        // t.chars().count(): Counts characters (Unicode scalar values).
+        let token_len = t.chars().count();
```
amitu, line 139: Since `t` is assumed ascii, we can assume `token_len` and `t.len()` are same. Or we can stop assuming that `t` is ascii. Treating `token_len` and `t.len()` as possibly different at the same time when we assume t is ascii is wrong. 

We currently do not have any non ascii tokens, so while I feel like saying lets relax ascii requirement, our scanner can handle emoji etc, yay, but since we will never be using it, its better to keep the contract. 

Can we achieve the effect you are trying to achieve by simply fixing this line: `self.s_index = self.index` as `self.s_index += count;`?

v0.5/fastn-p1/src/parser/mod.rs
```diff
@@ -46,7 +50,15 @@ macro_rules! tt {
...
-                fastn_p1::parser::p($source, $f, serde_json::json!($debug), $remaining);
+                fastn_p1::parser::p($source, $f, serde_json::json!($debug), $remaining, None);
+            };
+        }
+        /// test with index
+        #[allow(unused_macros)]
+        macro_rules! ti {
+            ($source:expr, $debug:tt, $remaining:expr, $chars:expr, $bytes:expr) => {
+                let index = fastn_p1::parser::scanner::Index::new($chars, $bytes);
+                fastn_p1::parser::p($source, $f, serde_json::json!($debug), $remaining, Some(index));
```
amitu, line 61:
```suggestion
            /// test with index
            ($source:expr, $debug:tt, $remaining:expr, $chars:expr, $bytes:expr) => {
                let index = fastn_p1::parser::scanner::Index::new($chars, $bytes);
                fastn_p1::parser::p($source, $f, serde_json::json!($debug), $remaining, Some(index));
```

amitu: No need to introduce a new macro name.

v0.5/fastn-p1/src/parser/section_init.rs
```diff
@@ -22,6 +22,7 @@ mod test {
         t!("-- foo: ", {"name": {"name": "foo"}}, " ");
         t!("-- foo: hello", {"name": {"name": "foo"}}, " hello");
         t!("-- integer foo: hello", {"name": {"name": "foo", "kind": "integer"}}, " hello");
+        ti!("-- integer héllo: foo", {"name": {"name": "héllo", "kind": "integer"}}, " foo", 17, 18);
```
amitu, line 25: Why do we have to check 17/18 here? The `assert` on `remaining` should be enough.

Arpita-Jaiswal: Basically `remaining` ensures if the value of `Scanner.index` is correct, but not `Scanner.s_index`.

amitu: Pass original source to `Scanner.remaining()`, and make it assert that remaining string generated by current char approach, and sub slice of original passed source are the same, and then it remains either.

amitu: `.remaining()` is `#[cfg(test)]` so this extra assertion / work is fine.

v0.5/fastn-p1/src/parser/mod.rs
amitu: Also create another clause in the macro so `remaining` is not needed, and it does an assertion that there is nothing remaining.

v0.5/fastn-p1/src/parser/scanner.rs
```diff
@@ -105,11 +105,13 @@ impl Scanner {
...
-            s.push(*c);
-        }
-        s
+        self.tokens[self.index..].iter().collect()
+    }
+
+    #[cfg(test)]
+    pub fn s_remaining(&self) -> String {
+        let input: String = self.tokens.iter().collect();
+        input[self.s_index..].to_string()
```
amitu, line 114: Nopes, don't create this. Make existing remaining() compute do the same and assert both methods produces same strings.

v0.5/fastn-p1/src/parser/scanner.rs
```diff
@@ -134,18 +136,28 @@ impl Scanner {
 
     // returns the span from current position to the end of token
     pub fn token(&mut self, t: &'static str) -> Option<fastn_p1::Span> {
-        let mut count = 0;
-        for char in t.chars() {
+        // Get the length of the token to match
+        let token_len = t.chars().count();
```
amitu, line 140: Keep my `mut count` solution. You are going through `t.chars()` twice. No need to do that.

Arpita-Jaiswal: But `if char != self.tokens[self.index + count] {` can overflow which `if self.index + token_len > self.size` took care of

amitu: Put an `&& self.index + count < self.size` there. 

Ideally we should look at generated assembly to see which one is shorter. But my point is let's try to really worry about performance this time.

v0.5/fastn-p1/src/parser/scanner.rs
```diff
@@ -39,6 +39,32 @@ impl Scanner {
...
+    /// Converts a given character count from the current index into the equivalent byte count.
+    fn char_count_to_byte_count(&self, char_count: usize) -> usize {
+        self.tokens[self.index..self.index + char_count]
+            .iter()
+            .map(|c| c.len_utf8()) // Get the byte length of each character
+            .sum() // Sum up the byte lengths
+    }
+
+    /// Advances the scanner's character and byte indices by a specified number of characters.
+    fn increment_index_by_count(&mut self, count: usize) {
```
amitu, line 51: Rename this to `increment_index_by`.

amitu (approved):

PR #2002, opened 2024-10-29
using iterator instead of vec

amitu:


v0.5/fastn-p1/src/parser/scanner.rs
```diff
@@ -52,7 +52,12 @@ impl Scanner<'_> {
         if index == self.index {
             return;
         }
-        while self.index > index {
+
+        // Don't like this. This will result in double allocations and moving offset again to index
+        self.chars = self.input.char_indices().peekable();
+        self.index = 0;
```
amitu, line 58: Would [`next_back()`](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html#tymethod.next_back) help?

PR #2004, opened 2024-10-29
Implement `kind` Parsing with Nested Arguments in `fastn 0.5`

Arpita-Jaiswal:
This PR finalizes the implementation of the `kind` function by fully implementing optional arguments parsing within the `Kind` structure. With this update, `kind` can now handle complex type-like structures with optional nested arguments enclosed in angle brackets (`<>`).

## Background and Previously Implemented Features

- **Qualified Identifier Parsing:**
The `kind` function initially parses a qualified identifier as the primary name of the `Kind` object.

- **Look-Ahead for Optional Arguments:**
After parsing the identifier, the function performs a look-ahead check for `<`. If `<` is not present, `kind` returns immediately with only the identifier as name, signaling that there are no arguments.

## Key Changes Introduced in this PR

- **Recursive Argument Parsing:**
  - If `<` is found, `kind` enters an argument parsing loop:
    - Parses each argument recursively as a `Kind`, allowing for nested types (e.g., `foo<bar<a>>`).
    - Separates arguments by commas `,` and terminates when a closing `>` is encountered, marking the end of the argument list.

- **Constructing the Kind Object:**
Constructs a `Kind` object with both `name` and `args` (if present).

## Example Cases

`foo<bar<k>>` ->
```json
{"name": {"module": "foo"}, "args": [{"name": {"module": "bar"}, "args": ["k"]}]}
```

`foo<a, b>` -> 
```json
{"name": {"module": "foo"}, "args": ["a", "b"]}
```

`foo<a, b<asd>, c, d>` ->
```json
{
    "name": {"module": "foo"},
    "args": [
        "a",
        {"name": {"module": "b"}, "args": ["asd"]},
        "c",
        "d"
    ]
}
```

### DIFF
--- a/v0.5/fastn-p1/src/utils.rs
+++ b/v0.5/fastn-p1/src/utils.rs
@@ -81,6 +81,10 @@ impl fastn_p1::Kind {
             return None;
         }
 
+        self.to_identifier_()
+    }
+
+    pub fn to_identifier_(&self) -> Option<fastn_p1::Identifier> {
         let m = self.name.module.as_ref()?;
         if !m.path.is_empty() {
             return None;
--- a/v0.5/fastn-p1/src/utils.rs
+++ b/v0.5/fastn-p1/src/utils.rs
@@ -81,10 +81,6 @@ impl fastn_p1::Kind {
             return None;
         }
 
-        self.to_identifier_()
-    }
-
-    pub fn to_identifier_(&self) -> Option<fastn_p1::Identifier> {
         let m = self.name.module.as_ref()?;
         if !m.path.is_empty() {
             return None;
--- a/v0.5/fastn-p1/Cargo.toml
+++ b/v0.5/fastn-p1/Cargo.toml
@@ -11,5 +11,7 @@ homepage.workspace = true
 
 [dependencies]
 serde.workspace = true
+
+[dev-dependencies]
 serde_json.workspace = true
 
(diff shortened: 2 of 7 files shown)


v0.5/fastn-p1/src/utils.rs
```diff
@@ -81,6 +81,10 @@ impl fastn_p1::Kind {
             return None;
         }
 
+        self.to_identifier_()
+    }
+
+    pub fn to_identifier_(&self) -> Option<fastn_p1::Identifier> {
```
amitu, line 87: Why? Is this performance optimisation or you do not want those checks?

Arpita-Jaiswal: `to_identifier_` is used here:
```rs
if let Some(v) = self.to_identifier_() {
    o.insert("name".into(), v.debug(source));
```
This helps in better JSON for testing the kind with arguments.

amitu: Why can't that use `to_identifier()`?

Arpita-Jaiswal: Because it has 'args' and there's a check in `to_identifier` which makes it return `None`

amitu: That check is to make sure you are not losing data. Which check if failing? Why are you ignoring that bit of data?

amitu: I think our `qualified_identifier` parser is wrong. 

From: 

```rs
        t!("foo", {"module": {"package": "foo"}});   // <-- wrong
        // t!("foo", {"module": "foo"}); <-- correct
```

Or we need some other term.

PR #2010, opened 2024-11-03
Document `f-script` for 0.5

siddhantk232:
This single file will contain every information that is needed to
implement `f-script`. This commit starts the documentation process.

It has covered:

- Existing f-script features
- New features that are coming in 0.5

This document needs more refinement and more details will come soon.


v0.5/fastn-lang/src/f-script/README.md
```diff
@@ -0,0 +1,221 @@
...
+string name:
+
+if name == "" {
+  "<default>"
+} else {
+  name
+}
+```
+
+The body contains a list of expressions. The last expression is evaluated and
```
amitu, line 16: We should introduce the new terminology. Not everything is an expression. 

````ftd
-- string foo():
string name:

string $foo: "hello";

if name.len() > 4 {
    foo = "bar";
}

if foo > name { foo } else { name }
```

The first and second statements in this function are not expression. Expressions are required only in the return position of a block, which is either an explicit return statement, or the last statement of the block.

v0.5/fastn-lang/src/f-script/README.md
```diff
@@ -0,0 +1,221 @@
...
+returned. There's no return keyword, if the function type is `void` then
```
amitu, line 17: > There's no return keyword.

This is wrong. We need early return statement.

amitu: We can note `void` is same as `()` in Rust.

v0.5/fastn-lang/src/f-script/README.md
```diff
@@ -0,0 +1,221 @@
...
+
+  name + (adj || " (Programmer)")
+}
+```
+
+- Control statements (`if..else`, `for`, `case`)
+
+```ftd
+{
+  let res = if name == "" { ;; nested f-script block
```
amitu, line 155: Instead of `f-script block`, let's use the terminology `block expression`, a la https://doc.rust-lang.org/beta/reference/expressions/block-expr.html (if is also f-script if, and let is also f-script let, there is no benefit of prefixing `f-script` to everything, and you are only prefixing it to `block`, making it seem like block is something special.

v0.5/fastn-lang/src/f-script/README.md
siddhantk232: I don't think we gain anything by making this distinction. Even in rust, `if` is not a statement but an expression. 

```rust
if name.len() > 4 {
    foo = "bar";
}
```

This `if` expression in rust would evaluate to `()` so this will compile.

```rust
if name.len() > 4 {
    4
} else {
    3
}
```

The above won't compile, the error would look something like:

```
error[E0308]: mismatched types
  --> test.rs:9:9
   |
8  | /     if 2 > 4 {
9  | |         4
   | |         ^ expected `()`, found integer
10 | |     } else {
11 | |         3
12 | |     }
   | |_____- expected this to be `()
```

Variable declarations can even be treated like expression that evaluate to `()` so you don't have to use the result of this expression. Evaluation of this expression would introduce a new symbol in your scope.


```rust
if foo > name { foo } else { name }
```

The above will either evaluate to `foo` or `name`. Since this is the last expression, it's value will be returned.

`match` and `for` can be thought of as expressions too. I am not trying to introduce "statement" unless I have to.

v0.5/fastn-lang/src/f-script/README.md
siddhantk232: This is done in https://github.com/fastn-stack/fastn/pull/2010/commits/2ab92c8f526a35437e6b0e5e55709bc1fb1d54d1

siddhantk232: > We can note void is same as () in Rust.
I'll ignore this for now.

v0.5/fastn-lang/src/f-script/README.md
siddhantk232: This is done in https://github.com/fastn-stack/fastn/pull/2010/commits/2ab92c8f526a35437e6b0e5e55709bc1fb1d54d1

v0.5/fastn-lang/src/f-script/README.md
amitu: Makes sense.

PR #2016, opened 2024-11-12
Merge `fastn-stack/fastn.com` website

siddhantk232:
This merges the history of [fastn-stack/fastn.com](https://github.com/fastn-stack/fastn.com/) into this repo under the `fastn.com` subfolder. Future deployments will be done using fastn upload.

This makes it easier to update docs along with feature additions to the framework.

### Other changes

- A gh workflow to optimize images in a PR.
- https://fifthtry.com/s/fastn is now **non-editable**
- [The Github (OIDC) setting](https://www.fifthtry.com/fastn/github/) now accepts changes from `fastn-stack/fastn`


amitu (approved):

.github/workflows/optimize-imgs.yml
```diff
@@ -0,0 +1,35 @@
...
+        uses: calibreapp/image-actions@main
+        with:
+          # The `GITHUB_TOKEN` is automatically generated by GitHub and scoped
+          # only to the repository that is currently running the action. By
+          # default, the action can’t update Pull Requests initiated from
+          # forked repositories.
+          # See:
+          #   - https://docs.github.com/en/actions/reference/authentication-in-a-workflow
+          #   - https://help.github.com/en/articles/virtual-environments-for-github-actions#token-permissions
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
```
amitu, line 35: This is kind of too much assumptions happening, but if we run this action on commit to main only, then this would not be needed. Nor would it be needed if we only do this for PRs created by people with access to this repo. Only for external fork PR this is needed. So if we keep this for PR, even then we can get rid of this line. 

I prefer running it on main tho, so if I do any commit bypassing PR workflow even then it runs. And this was it will run fewer times over all tho, as we may have many commits in PR and only one push to main per PR.

amitu:
https://www.fifthtry.com/fastn/github/

amitu:
@siddhantk232 this is odd, if I click on the link

.github/workflows/optimize-imgs.yml
siddhantk232: See https://github.com/fastn-stack/fastn/pull/2016/commits/e0f09c84156d522c2ffb87cdd1b0be475f62c90b. I've set this to run on push to `main` now. A separate PR with optimized images will be created if there's any output.

siddhantk232:
> > https://www.fifthtry.com/fastn/github/
> 
> @siddhantk232 this is odd, if I click on the link it gives error, but if I open it directly it opens.

Idk why this happens. I'll look into it. I've created an issue for this on our private issue tracker.

amitu (approved):

PR #2028, opened 2024-11-20
WIP: Parse `function_definition`s

siddhantk232:
Parse fastn function definitions. It's not finished in this PR.

It can parse function name, body, visibility (publica/private) and, return type. 

TODO: The function body itself needs to be parsed via the fscript parser which will turn it into a `fastn_resolved::FunctionExpression`. This will be done in another PR.


v0.5/fastn-section/src/parser/section_init.rs
```diff
@@ -7,11 +7,14 @@ pub fn section_init(
     scanner.skip_spaces();
     let name = fastn_section::parser::kinded_name(scanner)?;
     scanner.skip_spaces();
+    let function_marker = scanner.token("()");
```
amitu, line 10: I would first look for token `(`, skip space, and then look for `)`, so even `( )` etc can be parsed.

v0.5/fastn-section/src/parser/section_init.rs
```diff
@@ -26,6 +29,7 @@ mod test {
         t!("-- foo: hello", {"name": "foo"}, " hello");
```
amitu, line 29: Make jdebug return `{"section": "foo"}` here.

v0.5/fastn-unresolved/src/debug.rs
```diff
@@ -56,14 +56,92 @@ impl fastn_jdebug::JDebug for fastn_unresolved::AliasableIdentifier {
 
 impl fastn_jdebug::JDebug for fastn_unresolved::ComponentInvocation {
     fn debug(&self) -> Value {
-        todo!()
+        serde_json::json!({
+            "content": self.name.debug(),
+            "caption": self.caption.debug(),
```
amitu, line 61: Do these things conditionally, else test functions will become verbose. If we just had to do this then we would not need a custom trait and can use serde (in fact now that we do not depend on source for create value, we can maybe write an auto impl for everything implementing serde::Serialize so this kind of impls are not needed, and you only implement JDebug when you want to do "better than serde". 

The point of JDebug is to make the output as compact as possible so test cases look clean. The cleaner the test case, the more test cases we would write.

v0.5/fastn-unresolved/src/lib.rs
```diff
@@ -56,6 +56,12 @@ pub enum InnerDefinition {
     },
     Function {
         arguments: Vec<UR<Argument, fastn_resolved::Argument>>,
+        /// `None` means `void`. The `void` keyword is implied in fastn code:
+        /// ```ftd
+        /// -- foo(): ;; function with void return type
```
amitu, line 61: This is not correct, we use `-- void foo():`.

PR #2035, opened 2024-11-27
Refactor: custom `fastn_core::http::Response` for `fastn_core::serve`

siddhantk232:
`actix_web::HttpResponse` aka `fastn_core::http::Response` is not `Send` which makes it impossible to share it across threads. A direct downside of this is that we can't use a function which uses this type in a `tokio::spawn` because it required types to implement `Send`.

A custom `fastn_core::http::Response` is added which is `Send` and used by the `fastn_core::serve` function. A `fastn_core::serve_actix` is also introduced which is a direct replacement for the old `fastn_core::serve`.

This will allow us to, for example, use `fastn_core::serve` in a background task to update the cache.

### DIFF
--- a/fastn-core/src/commands/serve.rs
+++ b/fastn-core/src/commands/serve.rs
@@ -1,14 +1,15 @@
+use fastn_core::http::Response;
+
 #[tracing::instrument(skip_all)]
-fn handle_redirect(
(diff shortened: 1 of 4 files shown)


siddhantk232:
This is not needed anymore. Closing

PR #2056, opened 2025-01-27
Implicitly load `.wasm` modules from mounted apps

siddhantk232:
```ftd
-- fastn.app: Auth App
mount-point: /-/auth/
package: lets-auth.fifthtry.site
```

For the above configuration in a FASTN.ftd file. We allow the mounted app package to place some `.wasm` modules in its root directory.

So if the request is coming for: `/-/auth/api/*` and there is a "api.wasm" in the root of `lets-auth.fifthtry.site` then the request will be handled as if there was a wasm endpoint configured:

```ftd
-- fastn.url-mappings:

/-/auth/api/* -> wasm+proxy://.packages/lets-auth.fifthtry.site/api.wasm/*
```

That is, a request for `/-/auth/api/*`, will be handled by the `lets-auth.fifthtry.site/api.wasm` if it exists.

If the `.wasm` module does not exist then the typical flow of loading app's .ftd file is followed. This means that precedence is given to .wasm if it exists.

### DIFF
--- a/fastn.com/backend/app.ftd
+++ b/fastn.com/backend/app.ftd
@@ -1,0 +1,19 @@
+-- ds.page: Fastn app
+
+`-- fastn.app` allows you to mount a fastn package at some url of your fastn package.
+
+-- ds.code: FASTN.ftd
+lang: ftd
+
+\-- fastn.app: Auth App
+mount-point: /-/auth/
+package: lets-auth.fifthtry.site
+
+
+-- ds.markdown:
+
+The above snippet will mount contents of [lets-auth.fifthtry.site](https://lets-auth.fifthtry-community.com/) at the base url (`/-/auth/`) of your app.
+
+Visiting `/-/auth/` will load `index.ftd` of lets-auth.fifthtry.site if it's available.
+
+-- end: ds.page
--- a/fastn.com/backend/app.ftd
+++ b/fastn.com/backend/app.ftd
@@ -1,4 +1,4 @@
--- ds.page: Fastn app
+-- ds.page: `fastn` app
 
 `-- fastn.app` allows you to mount a fastn package at some url of your fastn package.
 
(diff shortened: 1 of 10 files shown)


siddhantk232:
This strictly follows https://github.com/fastn-stack/fastn/discussions/2051 except the part about giving precedence to `.ftd` files. `.ftd` files are only attempted if no `.wasm` file exists. In most cases, I expect the wasm file to be just called `api.wasm` since we have enough namespacing from the package name.

amitu:
> This strictly follows https://github.com/fastn-stack/fastn/discussions/2051 except the part about giving precedence to .ftd files.

Technically speaking this follow 2051 for `fastn.app` mounted apps, but not for the current package. Current package still requires explicit `wasm+proxy://` stuff. 

I wont block this on this technicality though, just noting it down, that we have inconsistent behaviour.

amitu (approved):

amitu (dismissed):

fastn.com/backend/app.ftd
```diff
@@ -0,0 +1,19 @@
+-- ds.page: Fastn app
```
amitu, line 1: We always refer to `fastn` as `fastn` in our docs, not "Fastn".

fastn.com/backend/wasm.ftd
```diff
@@ -0,0 +1,42 @@
+-- ds.page: WASM backends with [ft-sdk](https://github.com/fastn-stack/ft-sdk/)
+
+[ft-sdk](https://github.com/fastn-stack/ft-sdk/) is a rust crate that can be
```
amitu, line 3: We prefer `ft-sdk` instead of just "ft-sdk" (for crate names). Also it should be Rust, not rust.

siddhantk232: https://github.com/fastn-stack/fastn/pull/2056/commits/ffcdcfe2c1d2bb2134ac37150424cb5ee4daeb24 done in

amitu (approved):

PR #2057, opened 2025-01-28
`ftd.app-path` function

siddhantk232:
This function let's ftd files inside a mounted package construct urls
that are relative to the mountpoint of the said package.

This helps `fastn.app` mounts to easily construct paths without assuming
(or enforcing) their mountpoints.

Calling `ftd.app-path(path = /test/)` in an ftd file of a mounted app will return the path
prefixed with the `mountpoint` of the app.
                                                                                            
The `path` arg must start with a forward slash (/).

Calling this from any `.ftd` file of the root package will simply return the provided `path` argument.

                                                                                            
# Example
                                                                                            
```FASTN.ftd
-- import: fastn
                                                                                            
-- fastn.package: test
                                                                                            
-- fastn.app: Test
mountpoint: /app/
package: some-test-app.fifthtry.site
```
                                                                                            
```some-test-app.fifthtry.site/index.ftd
                                                                                            
-- ftd.text: $ftd.app-path(path = /test/)
```
                                                                                            
Visiting `/app/` in browser should render /app/test/

### DIFF
--- a/fastn-core/src/doc.rs
+++ b/fastn-core/src/doc.rs
@@ -28,8 +28,8 @@ fn cached_parse(
 pub fn package_dependent_builtins(
     pkg: &fastn_core::Package,
     req_path: &str,
-) -> Vec<(String, fastn_resolved::Definition)> {
-    vec![(
+) -> [(String, fastn_resolved::Definition); 1] {
+    [(
         "ftd#app-path".to_string(),
         fastn_core::host_builtins::app_path(pkg, req_path),
     )]
@@ -48,7 +48,7 @@ pub async fn interpret_helper(
     let doc = cached_parse(name, source, line_number)?;
 
     let builtin_overrides = package_dependent_builtins(&lib.config.package, lib.request.path());
-    let mut s = ftd::interpreter::interpret_with_line_number(name, doc, builtin_overrides)?;
+    let mut s = ftd::interpreter::interpret_with_line_number(name, doc, Some(builtin_overrides))?;
     lib.module_package_map.insert(
         name.trim_matches('/').to_string(),
         lib.config.package.name.to_string(),
--- a/fastn-core/src/doc.rs
+++ b/fastn-core/src/doc.rs
@@ -31,7 +31,7 @@ pub fn package_dependent_builtins(
 ) -> Vec<(String, fastn_resolved::Definition)> {
     vec![(
         "ftd#app-path".to_string(),
-        fastn_core::host_builtins::app_path(pkg, &req_path),
+        fastn_core::host_builtins::app_path(pkg, req_path),
     )]
 }
 
--- a/fastn-core/src/doc.rs
+++ b/fastn-core/src/doc.rs
@@ -25,7 +25,17 @@ fn cached_parse(
     fastn_core::utils::cache_it(id, C { doc, hash }).map(|v| v.doc)
 }
 
-#[tracing::instrument(skip_all)]
+pub fn package_dependent_builtins(
+    pkg: &fastn_core::Package,
+    req_path: &str,
+) -> Vec<(String, fastn_resolved::Definition)> {
+    vec![(
+        "ftd#app-path".to_string(),
+        fastn_core::host_builtins::app_path(pkg, &req_path),
+    )]
+}
+
+#[tracing::instrument(skip(lib))]
 pub async fn interpret_helper(
     name: &str,
     source: &str,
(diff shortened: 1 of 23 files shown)


ftd/src/interpreter/main.rs
```diff
@@ -92,6 +93,28 @@ impl InterpreterState {
...
+    fn new_with_expanded_builtins(
+        id: String,
+        builtins: impl IntoIterator<Item = (String, fastn_resolved::Definition)>,
+    ) -> InterpreterState {
+        let mut bag = ftd::interpreter::default::builtins().clone();
+
+        builtins.into_iter().for_each(|(name, def)| {
+            tracing::info!("Overriding builtin with def from fastn: {}", name);
+            bag.insert(name, def);
+        });
```
amitu, line 109: Why here and not in the builtin itself? We are using builtins from v0.5 code also, we will have to replicate this logic.

amitu:
@siddhantk232 also do the version number bump as we want to create a release right after merging this PR.

ftd/src/interpreter/main.rs
siddhantk232: This function is supposed to be invoked from `fastn-core` where we have context aware functions (that require `fastn_core::RequestConfig` and `fastn_core::Package`.

Moving this to `fastn-builtins` (which is shared with v0.5 and current) is no easy because we don't have access to `fastn-core`'s structs (we can't simply add them because circular dependency).

Other reason is that `fastn-builtins` is initialized once (`BUILTINS` static variable). Making the initialization logic take `RequestConfig` and related params that change on each request would make the program compute all the builtins per request. I decided to not do that and instead overwrite a few builtins per request. We still do a `.clone()` of the builtins (which we did before this PR also), this is one part that can be improved later to save some memory.

ftd/src/interpreter/main.rs
```diff
@@ -92,6 +93,28 @@ impl InterpreterState {
         }
     }
 
+    /// [InterpreterState] with stub functions replaces with fastn_core::Package dependant
+    /// functions.
+    /// You pass real functions using the `builtins` parameter.
+    #[tracing::instrument(skip(builtins))]
+    fn new_with_expanded_builtins(
+        id: String,
+        builtins: impl IntoIterator<Item = (String, fastn_resolved::Definition)>,
```
amitu, line 102: I usually prefer `builtins: &[(String, fastn_resolved::Definition)]`. While yours is more generic, this is slightly less to type (and relies of the fact that it will be a vector or array to begin with, not a general purpose iterator).

fastn-core/src/doc.rs
```diff
@@ -25,7 +25,17 @@ fn cached_parse(
     fastn_core::utils::cache_it(id, C { doc, hash }).map(|v| v.doc)
 }
 
-#[tracing::instrument(skip_all)]
+pub fn package_dependent_builtins(
+    pkg: &fastn_core::Package,
+    req_path: &str,
+) -> Vec<(String, fastn_resolved::Definition)> {
```
amitu, line 31: Instead of returning a Vec you can return a fixed size array (and change function definition when the array length changes, eg when you add more items, to try to avoid one malloc, as arrays would be created on the stack.

siddhantk232: done in https://github.com/fastn-stack/fastn/pull/2057/commits/897fc62b9098dcf1ef6cca4629f8fd5679501c7b

ftd/src/interpreter/main.rs
siddhantk232: done in: https://github.com/fastn-stack/fastn/pull/2057/commits/897fc62b9098dcf1ef6cca4629f8fd5679501c7b

ftd/src/interpreter/main.rs
```diff
@@ -99,14 +99,16 @@ impl InterpreterState {
     #[tracing::instrument(skip(builtins))]
     fn new_with_expanded_builtins(
         id: String,
-        builtins: impl IntoIterator<Item = (String, fastn_resolved::Definition)>,
+        builtins: Option<[(String, fastn_resolved::Definition); 1]>,
```
amitu, line 102: You do not need option, you can pass an empty array. Also receiving functions should not accept array but slice, `builtins: &[(String, fastn_resolved::Definition)]`.

amitu: Arrays have fixed size. Vector has dynamic size. Slice has size unknown at compile time and either array or vec can deref to a slice.

amitu: When generating data, vec or array is needed, but when consuming a list like, slice should be preferred as it can accept both vec or array.

siddhantk232: Using a fixed sized array (owned) avoids `.clone()` as slice is borrowed:

```rust
    fn new_with_expanded_builtins(
        id: String,
        builtins: &[(String, fastn_resolved::Definition)],
    ) -> InterpreterState {
        let mut bag = ftd::interpreter::default::builtins().clone();

          for (name, def) in builtins {
              tracing::info!("Overriding builtin with def from fastn: {}", name);
              bag.insert(name.clone(), def.clone()); // This has  .clone() if we use a slice!
          }
     
        InterpreterState {
            id,
            bag,
            ..Default::default()
        }
    }
```

I think it's okay to move in the array as it's not supposed to be used anywhere. Ideally, the `bag` should store references (but that will involve a bunch of lifetimes that I don't wanna deal with right now).

I can change this to use slices if you like but I want you to know about the `.clone()` that'll then have to be used. Should I do it anyway?

amitu: Trying to avoid clone makes sense, I think in this instance what you have done is better.

amitu (approved):

PR #2059, opened 2025-01-30
Send `fastn.app` mountpoints to .wasm modules via `x-fastn-app-mounts` header

siddhantk232:
The header value is a json object that contains the mountpoints of `fastn.app`s

If in FASTN.ftd, we have:

```ftd
-- import: fastn
-- fastn.package: hello-world

-- fastn.app: Auth App
package: lets-auth.fifthtry.site
mount-point: /-/auth/

-- fastn.app: Let's Talk App
package: lets-talk.fifthtry.site
mount-point: /talk/
```

Then the value will be a json string:

```json
{ "lets-auth": "/-/auth/", "lets-talk": "/talk/" }
```

NOTE: `lets-auth.fifthtry.site` and `lets-talk.fifthtry.site` are required to be a system
package. The names `lets-auth` and `lets-talk` are taken from their `system` field.

A check is also added that restricts any package from getting mounted more than once. Mounted apps also have to be a system package.

### DIFF
--- a/fastn-core/src/commands/serve.rs
+++ b/fastn-core/src/commands/serve.rs
@@ -587,9 +587,20 @@ async fn handle_endpoints(
     tracing::info!("url: {}", url);
 
     if url.starts_with("wasm+proxy://") {
+        let app_mounts = match config.package.app_mounts() {
+            Err(e) => return Some(Err(e)),
+            Ok(v) => v,
+        };
+
         return match config
             .ds
-            .handle_wasm(url, req, endpoint.mountpoint.to_string(), session_id)
+            .handle_wasm(
+                url,
+                req,
+                endpoint.mountpoint.to_string(),
+                app_mounts,
+                session_id,
+            )
             .await
         {
             Ok(r) => Some(Ok(to_response(r))),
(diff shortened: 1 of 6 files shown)


fastn-core/src/commands/serve.rs
```diff
@@ -587,9 +587,20 @@ async fn handle_endpoints(
     tracing::info!("url: {}", url);
 
     if url.starts_with("wasm+proxy://") {
+        let app_mounts = match config.package.app_mounts() {
```
amitu, line 590: Make this `config.app_mounts()`.

siddhantk232: done in https://github.com/fastn-stack/fastn/pull/2059/commits/b74a3d1983a1ec4e5b50157d1f73e8810a1b400b

amitu: In case you are curious, we did this because "app_mounts" only makes sense for "top level package", which is `config.package`, but the `Package` type is used in a bunch of other places as well, and if someone called `Package.app_mounts()` for non "top level package" it would be garbage, so we want that to be compile time impossibility, by having `Config.app_mounts()` which will always make sense.

amitu (approved):

PR #2060, opened 2025-01-31
Feat/ftd main package

siddhantk232:
# TODO

- [x] `ftd.main-package` variable
- [ ] Add `app` param to `ftd.app-path`
- [ ] Update docs of `ftd.app-path`
- [ ] Document `ftd.main-package`

### DIFF
--- a/fastn-core/src/host_builtins.rs
+++ b/fastn-core/src/host_builtins.rs
@@ -22,7 +22,7 @@
 ///
 /// Visiting `/app/` in browser should render /app/test/
 #[inline]
-pub fn app_path(pkg: &fastn_core::Package, req_path: &str) -> fastn_resolved::Definition {
+pub fn app_path(pkg: &fastn_core::Package, req_path: &str) -> (String, fastn_resolved::Definition) {
     let prefix = pkg
         .apps
         .iter()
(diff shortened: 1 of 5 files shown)


fastn-core/src/host_builtins.rs
```diff
@@ -57,5 +58,33 @@ pub fn app_path(pkg: &fastn_core::Package, req_path: &str) -> fastn_resolved::De
...
+    });
+
+    (name, def)
+}
+
+/// Ftd string variable that holds the name of the package.
+///
+/// Useful to determine if the package is run standalone or as a dependency:
+#[inline]
+pub fn main_package(pkg: &fastn_core::Package) -> (String, fastn_resolved::Definition) {
```
amitu, line 70: This should also take `fastn_core::Config` as argument, for the same reason I discussed yesterday.

amitu: Also maybe you can make `impl Config { fn main_package() {} }` so it shows up in code completion etc, which has better discovery than scattered utility methods.

amitu:
In the task list add docs as well.

siddhantk232:
Closing in favor of #2061

PR #2061, opened 2025-01-31
`ftd.main-package` (variable) and `ftd.is_app_mounted` (function)

siddhantk232:
Demo video: https://youtu.be/IrfsVCy7Y0Y

# TODO

- [x] `ftd.main-package` variable
- [x] Add `app` param to `ftd.app-path`
- [x] `ftd.is_app_mounted` function
- [x] Update docs of `ftd.app-path`
- [x] Document `ftd.main-package`
- [x] Document `ftd.is_app_mounted`
- [x] Rename `ftd.app-path` -> `ftd.app-url`

### DIFF
--- a/fastn-builtins/src/lib.rs
+++ b/fastn-builtins/src/lib.rs
@@ -1037,52 +1037,6 @@ pub fn default_bag() -> indexmap::IndexMap<String, fastn_resolved::Definition> {
             })
         ),
         (
-            "ftd#app-path-ex".to_string(),
-            fastn_resolved::Definition::Function(fastn_resolved::Function {
-                name: "ftd#app-path-ex".to_string(),
-                return_kind: fastn_resolved::KindData {
-                    kind: fastn_resolved::Kind::string(),
-                    caption: false,
-                    body: false,
-                },
-                arguments: vec![
-                    fastn_resolved::Argument {
-                        name: "path".to_string(),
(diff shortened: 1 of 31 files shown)


fastn-builtins/src/lib.rs
```diff
@@ -1036,6 +1036,52 @@ pub fn default_bag() -> indexmap::IndexMap<String, fastn_resolved::Definition> {
                 external_implementation: true
             })
         ),
+        (
+            "ftd#app-path-ex".to_string(),
```
amitu, line 1040: I was thinking you make the `app` a required field and only expose one function.

siddhantk232: I've removed `app-path-ex` form builtins. `app_path_ex` in ftd.js file will be internally called. The `app` param is still optional. If it's not passed then we take the current app from the context (i.e, if the file is present in lets-auth package then app = "lets-auth"). Do you want me to make this explicit? I kinda like this implicitness, especially when you consider that, as an app author, your code will be filled with such function calls where you're constructing urls for your app but you still have to pass your app name every time.

amitu: No, the behaviour you describe is perfect, I like optional app as well, and the general pattern of optional function values (and optional component arguments etc) for fastn ecosystem.

amitu (approved):

PR #2062, opened 2025-01-31
support `ftd.json`

amitu:
We want people to able to return JSON data from ftd files as well:

```
-- ftd.json:
sign-up-url: $lets-talk.sign-up-url
```

<details><summary>current bug</summary>

```
curl http://127.0.0.1:8000/-/auth/config/
{"sign-up-url":{"String":{"text":"$lets-talk.sign-up-url"}}}
```

</details>


amitu:
Good job.
```ftd
-- integer x: 20
-- record foo:
integer x:
string y:
boolean z:

-- foo f:
x: $x
y: hello
z: false

-- ftd.json:
yo-data: $x
f: $f
name: Sam
inline-integer-is-string: 10
inline-bool: false
url: $lets-auth.sign-up-url
```
My understanding was inline values would be strings, I was expecting 
```json
{"f":{"x":20,"y":"hello","z":false},"inline-bool":false,"inline-integer-is-string":10.0,"name":"Sam","url":"/-/auth/signup/","yo-data":20}{"f":{"x":20,"y":"hello","z":false},"inline-bool":"false","inline-integer-is-string":"10","name":"Sam","url":"/-/auth/signup/","yo-data":20}
```
But got :
```json
{"f":{"x":20,"y":"hello","z":false},"inline-bool":false,"inline-integer-is-string":10.0,"name":"Sam","url":"/-/auth/signup/","yo-data":20}{"f":{"x":20,"y":"hello","z":false},"inline-bool":false,"inline-integer-is-string":10.0,"name":"Sam","url":"/-/auth/signup/","yo-data":20}
```
Observe the value of `"inline-integer-is-string"` and `"inline-bool"` fields in json.

I am quite torn on what to do here, this seems like better behaviour than what I was going to get, but then this brings to YML territory where things can accidentally read as integer or bool when you meant them as strings for whatever reason. 
```ftd
-- string k: 10

-- ftd.json:
k: $k
```
You can force string etc by giving it a named variable, so maybe this is ok. But should we go for behaviour a: correct (always string) and overwrite using explicit variable, or b: "convenient" (no need to create explicit variable for obviously integer and booleans, and overwrite them to string using explicit variable.

One reason to pick convenient is its what in the PR already, no work is required. I am thinking of merging the PR, and we can review this decision close to 0.5 release.

PR #2070, opened 2025-02-08
Etag support in `fastn update`

amitu:
Both Github, and not FifthTry support ETag for zip downloads, which we should use to optimise downloads. We are going to store the Etag in `.packages/<package-name>/..etag`.


fastn-core/fbt-tests/19-offline-build/cmd.p1
```diff
@@ -11,14 +11,12 @@ Processing fastn-community.github.io/business-card-demo/assets/ipsum-logo-dark.s
 Processing fastn-community.github.io/business-card-demo/assets/ipsum-logo.svg ... done in <omitted>
 Processing fastn-community.github.io/business-card-demo/ ... Processing fastn-community.github.io/business-card/assets/fastn-badge-white.svg ... done in <omitted>
 Processing fastn-community.github.io/business-card/assets/fastn-badge-white-dark.svg ... done in <omitted>
-Processing fastn-community.github.io/business-card/assets/download-hover.svg ... done in <omitted>
```
amitu, line 14: Why is this happening?

PR #2076, opened 2025-02-13
Added `ftd.response`

Arpita-Jaiswal:
[Discussion 2066: `ftd.response` with rudimentary variable interpolation](https://github.com/orgs/fastn-stack/discussions/2066)

- [x] basic feature
- [x] content-type default value to be text/html
- [x] status code support
- [x] header support

These two we can do in future PR.

- [ ] variable interpolation in body


ftd/src/interpreter/main.rs
```diff
@@ -1349,6 +1349,85 @@ impl Document {
         Ok(None)
     }
 
+    // Returns (response, content-type, status, headers)
+    #[allow(clippy::type_complexity)]
+    pub fn get_response(
+        &self,
+    ) -> ftd::interpreter::Result<Option<(String, String, u16, fastn_resolved::Map<String>)>> {
```
amitu, line 1356: Should return `actix_web::http::StatusCode` instead of `u16`.

ftd/src/interpreter/main.rs
```diff
@@ -1349,6 +1349,85 @@ impl Document {
...
+                    );
+                }
+            };
+
+            let status_code = match v.get_interpreter_value_of_argument("status-code", &tdoc)? {
+                Some(val) => val.integer(self.name.as_str(), v.line_number)?,
+                None => 200,
+            };
+
+            if !(100..1000).contains(&status_code) {
```
amitu, line 1393: Instead of explicit logic here, let's use `actix_web::http::StatusCode::from_u16`. We will have to do both checks, check that it is less than u16::MAX, if so cast it as u16, and then use from_u16 to see if its valid.

ftd/src/interpreter/main.rs
amitu: We cannot do this because then we will have to add actix_web dependency in ftd crate.

PR #2100, opened 2025-02-25
`export: *`

Arpita-Jaiswal:
This PR introduces support for `export: *`, allowing all symbols from an imported module to be re-exported. However, this can only be used once per file in its current implementation.

## Next Steps
A future PR will extend support for multiple `export: *` statements within the same file.

## Example Usage
```ftd
-- import: bar
export: *
```


amitu:
@Arpita-Jaiswal  does it work for our use case, like `export: *` should have lower precedence than symbols defined in the same file after the `export: *` line?

Like if `ds.ftd` has `heading-large` and `heading-medium`, and we use `-- import: ds\nexport: *` in `myds.ftd`, but define our own `heading-large`, then `myds.heading-large` have to be the one defined in `myds`, but `myds.heading-medium` should be the one defined in `ds`?

I can test it but it would be easier for you to tell me.

siddhantk232:
> @Arpita-Jaiswal does it work for our use case, like `export: *` should have lower precedence than symbols defined in the same file after the `export: *` line?
> 
> Like if `ds.ftd` has `heading-large` and `heading-medium`, and we use `-- import: ds\nexport: *` in `myds.ftd`, but define our own `heading-large`, then `myds.heading-large` have to be the one defined in `myds`, but `myds.heading-medium` should be the one defined in `ds`?
> 
> I can test it but it would be easier for you to tell me.

could write a test for this.

amitu:
> could write a test for this.

@siddhantk232 sometimes in language its okay to omit some words, sometimes it matters. I am assuming the word you omitted was "I can...", while knowing you probably meant "You can...".

siddhantk232:
> > could write a test for this.
> 
> @siddhantk232 sometimes in language its okay to omit some words, sometimes it matters. I am assuming the word you omitted was "I could...", while knowing you probably meant "You could...". :-P

I didn't mean to be ambiguous, my bad. "You should write a test for this" is what I meant.

amitu:
@Arpita-Jaiswal can you create some JSON example, so its easy to see:

`a.ftd`

```ftd
-- integer x: 10
-- integer y: 20
```

`b.ftd`

```ftd
-- import: a
export: *

-- integer y: 30
```

`j.ftd`

```ftd
-- import: b
-- ftd.json:
x: $b.x
y: $b.y
```

Should show:

```json
{
    "x": 10,
    "y": 30
}
```

amitu (dismissed):

PR #2116, opened 2025-03-27
Add `autofocus` attribute to `ftd#text-input`

siddhantk232:
```ftd
-- ftd.text-input:
placeholder: Enter search query...
autofocus: true ;; this is NEW!
```

This newly added attribute is also used in `fastn.com/search.ftd`'s text-input. The default value is also set to `ftd.empty` instead of "Home".


amitu:
From the copilot review:

> [nitpick] Ensure that the naming and usage of autofocus here aligns with the rest of the codebase and matches the HTML attribute convention (all lowercase 'autofocus').

@siddhantk232 do you think it should be `auto-focus: true`, or you prefer `autofocus: true` only?

siddhantk232:
> From the copilot review:
> 
> > [nitpick] Ensure that the naming and usage of autofocus here aligns with the rest of the codebase and matches the HTML attribute convention (all lowercase 'autofocus').
> 
> @siddhantk232 do you think it should be `auto-focus: true`, or you prefer `autofocus: true` only?

I picked the HTML attribute name `autofocus` which, in my opinion, goes well with `multiline` and `placeholder` that are already supported by `ftd#text-input`. There are other attributes too that mostly are in kebab-case. Not sure what's the right thing here, I'd stick with `autofocus` for now.

amitu (approved):

siddhantk232:
Updated the old Changelog.md file, also update docs of `ftd.text-input`. @amitu

amitu (approved):

PR #2117, opened 2025-03-29
Changing set function for MutableVariable with value of type RecordInstance

ayushdeolasee:
Changing incorrect function to deal with set function for a MutableVariable that holds a RecordInstance as a value.


amitu:
@ayushdeolasee if you modify js files, the fbt tests fail, to fix them you have to run `cd fastn-core && fbt -f`.

ayushdeolasee:
Oh okay, will do that

ayushdeolasee:
@amitu, the previous tests which were failing were fixed by running fbt -f, but right now the test `js::ftd_test_helpers::fastn_js_test_all` is failing, is there anything else that needs be done for this test?

amitu:
@ayushdeolasee in [the workflow file](https://github.com/fastn-stack/fastn/blob/main/.github/workflows/tests-and-formatting.yml#L89-L93), there are a bunch of commands, try to see if any of them work.

ayushdeolasee:
Using this fix was creating a stack overflow error in the tests, so it was better to access the `RecordInstance` within the `MutableVariable` and call the `set` function on the `RecordInstance` in the use case of the calendar widget.

amitu:
@ayushdeolasee can you share some code snippet here for the workaround you mentioned?

PR #2129, opened 2025-05-09
Fix: Consider sitemap items when resolving imports 

siddhantk232:
```ftd
;; guide/install.ftd
-- string test: some val
-- ftd.text: $test
```

```ftd
;; FASTN.ftd
...

-- fastn.sitemap:

\# Installation: /install/
  document: guide/install.ftd
```

This will fail to build, reason:

fastn is trying to resolve `$test` in `guide/install.ftd` and in doing
so, it's resolving the current module with id: `/install/` but this
fails as the id translates to a `install.ftd` (or `/install/index.ftd`)
in project dir. The correct id should be `guide/install/` (document
attributed in sitemap definition).

This is fixed by looking into sitemp entries if available and changing
the id before trying to do `package.resolve_by_id`.

NOTE: The build fails only with `fastn build`. `fastn serve` works fine without this change.

### DIFF
--- a/fastn-core/src/doc.rs
+++ b/fastn-core/src/doc.rs
@@ -138,7 +138,8 @@ pub async fn interpret_helper(
 }
 
 // source, foreign_variable, foreign_function
-#[tracing::instrument(skip_all)]
+#[allow(clippy::type_complexity)]
+#[tracing::instrument(skip(lib, _state))]
 pub async fn resolve_import_2022(
     lib: &mut fastn_core::Library2022,
     _state: &mut ftd::interpreter::InterpreterState,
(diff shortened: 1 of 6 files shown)


fastn-core/src/doc.rs
```diff
@@ -58,6 +58,7 @@ pub async fn interpret_helper(
     loop {
         match s {
             ftd::interpreter::Interpreter::Done { document: doc } => {
+                tracing::info!("done");
```
amitu, line 61: Can we remove these?

amitu (approved):

fastn-core/src/doc.rs
siddhantk232: It's useful in debugging. Tells you which state the interpreter is currently in. This is all hidden behing `fastn --trace ...` anyway.

If you still want these removed, I'll remove them in the next commit (in another PR).

PR #2131, opened 2025-05-16
Fastn new book

deepti-mahadeva:


amitu:
@deepti-mahadeva since we use rebase merge strategy, if you care a banrch for one PR and that gets merged (via rebase strategy), you should not continue to push in same branch as that leads to conflicts (and bad commit messages, the ones that are already merged show up again). We can either stop using rebase strategy, which leads to very non linear git histories, all the merge commits etc, or we follow the "always create a new branch from a branch off of main after you have pulled your main to latest (so it includes your past PR commits) discipline".

deepti-mahadeva:
> @deepti-mahadeva since we use rebase merge strategy, if you create a branch for one PR and that gets merged (via rebase strategy), you should not continue to push in same branch as that leads to conflicts (and bad commit messages, the ones that are already merged show up again). We can either stop using rebase strategy, which leads to very non linear git histories, all the merge commits etc, or we follow the "always create a new branch from a branch off of main after you have pulled your main to latest (so it includes your past PR commits)" discipline.

@amitu I will create new branch

PR #2135, opened 2025-05-25
Build static fastn for x86_64-linux-gnu

siddhantk232:
The CI is setup to build static `fastn` binary. This is done by statically linking with the glibc on linux.

This is tested on NixOS with the `pkgs.glibc.static` package. From stackoverflow[1], it's equivalent on ubuntu is `libc6-dev` so this should work on github CI too.

An attempt to compile using `musl-gcc` was also made but it did not work because crates `rusqlite` (because libsqlite-sys) and `quickjs` failed to link with musl's libc.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -15,7 +15,7 @@ jobs:
     # using the oldest available ubuntu on github CI to provide maximum compatibility with glibc versions
     # update RELEASE_TEMPLATE with the glibc version
     # on ubuntu-22.04, glibc version is 2.35
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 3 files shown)


.github/workflows/create-release.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     # using the oldest available ubuntu on github CI to provide maximum compatibility with glibc versions
     # update RELEASE_TEMPLATE with the glibc version
     # on ubuntu-22.04, glibc version is 2.35
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
```
amitu, line 18: Use specific version so CI does not break when GitHub upgrades their systems.

amitu (approved):

PR #2140, opened 2025-06-08
wip: post is working

amitu:
When a `wasm+proxy://` url is used in an `http` processor, the POST request body were not sent. This is fixed in this PR. Support for tuple style POST body params is also added to `ftd.submit_form`, this works similar to the `ftd.http` function:

```ftd
ftd.submit_form(
    action_url,
    ("concept-url", concept_url), ;; ignore server sent errors by not providing a third element
    ("status", status, error) ;; server sent error for "status" will be stored in the `error` variable
)
```

The third element of the tuple is an optional mutable variable defined as:

```ftd
-- optional string $status-error:
```

Note that the use of `optional` is necessary as using just the `string`
type would issue warning and will not get updated from js.

- fix: Send POST body to a wasm+proxy:// url.
- feat: `ftd.submit_form` support tuple style args for POST body params.
- cleanup: remove some unused rust functions

The associated ft_sdk [PR](https://github.com/fastn-stack/ft-sdk/pull/65) is not necessary to work with changes in this PR. It can be merged later.


fastn-core/src/config/mod.rs
```diff
@@ -1201,22 +1201,22 @@ impl Config {
...
+    //     let package = fastn_core::Package::new(package_name);
+    //     let root = self.get_root_for_package(&package);
+    //     let package_fastn_path = root.join("FASTN.ftd");
+    //     let doc = self
+    //         .ds
+    //         .read_to_string(&package_fastn_path, session_id)
+    //         .await?;
+    //     let lib = fastn_core::FastnLibrary::default();
+    //     Ok(fastn_core::doc::parse_ftd("fastn", doc.as_str(), &lib)?)
+    // }
```
siddhantk232, line 1219: why comment these? either remove them or keep it uncommented so that it shows in the IDE autocomplete when writing code.

PR #2143, opened 2025-06-10
Better release notes and support for form level errors in `ftd.submit_form`

siddhantk232:
In addition to doing this:

  ```ftd
  ftd.submit_form(
      action_url,
      ("concept-url", concept_url),
      ("status", status, error)
  )
  ```

You can now also do:

  ```ftd
  ftd.submit_form(
      (action_url, form_error), ;; NOTICE THIS!
      ("concept-url", concept_url),
      ("status", status, error)
  )
  ```

This allows the server to send an error string for the key "all" whose
value will be saved in the `form_error` variable.

NOTE: with this change the key "all" is reserved now. This means that
you can not send a key "all" in your HTTP requests.

The release notes are fetched from our CHANGELOG.md file now.


amitu (approved):
Awesome job on ensuring by developer is notified when they mistakenly use all as a key for regular field.

PR #2144, opened 2025-06-11
Support requests to wasm modules in http processor

siddhantk232:
## Handle wasm modules in mountpoints of apps

TLDR: the following code works now:

```ftd
;; FASTN.ftd
-- import: fastn

-- fastn.package: test

-- fastn.dependency: lets-teach.fifthtry.site

-- fastn.app: Lets Teach App
mount-point: /teach/
package: lets-teach.fifthtry.site
```

```ftd
;; index.ftd

-- some-record page-data:
$processor$: pr.http
method: post
url: /teach/backend/hello/
```

`/teach/backend/hello/` maps to a "hello" handler defined in the backend.wasm module which is present at `lets-teach.fifthtry.site/backend.wasm`.

<details><summary>Implementation details:</summary>

We check if the url is part of a mounted app in the current package. If
it is, then we try to create a wasm+proxy:// url for implicit wasm
modules present in the app's directory. If the implicit wasm module does
**not** exists, we simply bail by creating a full url (prefixed with
http/https) and let the fastn server logic handle it when the http processors
calls the full url.
</details>

## Handle package local relative urls

This is useful in situations like the following:

```ftd
;; index.ftd
-- some-record-name d:
$processor$: processors.http
url: /local/
```

```ftd
;; local.ftd

-- ftd.json
id: 1
name: Someone
```

Instead of returning an error in case of no resolution of the url. We
simply return the full url (path prefixed with scheme+host) in hopes
that fastn server logic will handle it.

The resolved will convert url `/local/` -> `http://127.0.0.1:800/local/`
(depends on running env) which is where the http request from the
processor will be sent and fastn server will realise that it's a local
file and will return the correct json response.

**NOTE**: The http processor enhancements don't work with `fastn build`. Another PR should at least add a better error message when the processors fails and should show hints (using full urls would make it work with `fastn build`) to fix it.

## fix: Wrong filename in the step to extract changelog

This will not let the CI generate the "What's Changed" section.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -176,7 +176,7 @@ jobs:
             found && /^## [0-9]{2}/ { exit }
             found && /^### malai / { exit }
             found { print }
-          ' CHANGELOG.md | sed "1s/.*/# What's Changed/" >> .github/RELEASE_TEMPLATE.md
+          ' Changelog.md | sed "1s/.*/# What's Changed/" >> .github/RELEASE_TEMPLATE.md
       - uses: ncipollo/release-action@v1
         with:
           artifacts: "~/download/windows/fastn_windows_x86_64.exe,~/download/windows/fastn_setup.exe,~/download/macos/fastn_macos_x86_64,~/download/linux/fastn_linux_musl_x86_64"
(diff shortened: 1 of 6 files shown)


amitu:
Nice touch on the /ftd-file/ stuff!

PR #2152, opened 2025-06-17
fix: Server render meta tags if request is from a bot

siddhantk232:
A bot for example is Linkeding | facebook | discord etc. We check if the
request is coming from them, then we take the generated js and execute
it on the server (ssr) to create those meta tags and then inject them
into the output html.

Normal browser request don't take the perf hit of running js on the
server.

This requires no change in user code. Existing use of `ftd#document`
should just work.

### DIFF
--- a/fastn-core/src/package/package_doc.rs
+++ b/fastn-core/src/package/package_doc.rs
@@ -572,23 +572,10 @@ pub(crate) async fn read_ftd_2023(
         )
     } else {
         let (ssr_body, meta_tags) = if config.request.is_bot() {
-            let ssr_res = fastn_js::ssr_with_js_string(
+            fastn_js::ssr_with_js_string(
                 &package_name,
                 format!("{js_ftd_script}\n{js_document_script}").as_str(),
-            )?;
-
-            assert_eq!(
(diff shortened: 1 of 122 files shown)


fastn-core/src/package/package_doc.rs
```diff
@@ -571,20 +571,34 @@ pub(crate) async fn read_ftd_2023(
             format!("{js_ftd_script}\n{js_document_script}").as_str(),
         )
     } else {
-        let ssr_body = if config.request.is_bot() {
-            fastn_js::ssr_with_js_string(
+        let (ssr_body, meta_tags) = if config.request.is_bot() {
+            let ssr_res = fastn_js::ssr_with_js_string(
```
amitu, line 575: I would prefer if this function returned a string tuple instead of vec, as these checks we have performed as internal details of this function.

amitu (dismissed):

fastn-core/src/package/package_doc.rs
siddhantk232: Done in

amitu (approved):

PR #2158, opened 2025-06-23
Implementing fastn-xtask library crate

ayushdeolasee:
Creating a library crate that can be used to replace the scripts folder currently used to execute commands for fastn projects.


fastn-xtask/src/build_wasm.rs
```diff
@@ -0,0 +1,95 @@
+use std::env;
```
amitu, line 1: We do not use `use` statement. Use full path everywhere.

fastn-xtask/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+        Some("update-www") => update_www::update_www().map_err(|e| e.to_string())?,
+        Some("run-www") => run_www::run_www().map_err(|e| e.to_string())?,
+        Some("update-template") => update_template::update_template().map_err(|e| e.to_string())?,
+        _ => print_help(),
+    }
+
+    Ok(())
+}
+
+fn print_help() {
```
amitu, line 52: Make this pub and accept an optional str arg that may contain help for other commands the user may have defined. 

This is for if they do not want to use our main function. 

Actually do this, in one of our repositories do not use main and create an extra command. And ensure the story there is also pleasant.

fastn-xtask/src/new_app.rs
```diff
@@ -0,0 +1,185 @@
+pub async fn new_app(app_name: &str) -> fastn_core::Result<()> {
```
amitu, line 1: This won't be in this crate. After I have created an app only will xtask will be present. Fastn new -t or fastn template etc will create new app, not this crate.

fastn-xtask/src/run_template.rs
```diff
@@ -0,0 +1,86 @@
+use std::env;
```
amitu, line 1: There is a lot of duplication in these run_* methods, they all do the same but operate on a different folder. So create some run helper and call it from the three run_* functions.

fastn-xtask/src/update_template.rs
```diff
@@ -0,0 +1,60 @@
+use std::env;
```
amitu, line 1: Same here. Create update helper.

fastn-xtask/src/run_ui.rs
```diff
@@ -0,0 +1,59 @@
...
+use std::fs;
+
+pub fn run_ui() -> fastn_core::Result<()> {
+    println!("Running UI");
+
+    let current_dir = env::current_dir().map_err(|e| {
+        fastn_core::Error::GenericError(format!("Failed to get current directory: {}", e))
+    })?;
+
+    let re = Regex::new(r".*\.fifthtry\.site$").unwrap();
```
amitu, line 13: Don't use regex here. It's a heavy crate and all you need is ends_with check.

fastn-xtask/src/update_template.rs
```diff
@@ -1,44 +1,23 @@
-use std::env;
-use std::process::Command;
-use regex::Regex;
-use std::fs;
-
 pub fn update_template() -> fastn_core::Result<()> {
     println!("Updating template");
 
-    let current_dir = env::current_dir().map_err(|e| {
+    let template_dir = crate::helpers::find_directory(
```
amitu, line 4: We do not use `crate`, we give the package a name, e.g., this one should be called `fastn_xtask` and the line should read `fastn_xtask:helpers::find_directory()`. You will have to add following to `fastn-xtask/src/lib.rs`

```rs
extern crate self as fastn_xtask;
```

amitu:
@ayushdeolasee you should resolve a conversation once you have done it. Ideally with a link to commit where you have implemented the requested change.

amitu (approved):

Issue #2161, opened 2025-06-27
Changing value of cookie `fastn-lang` does not change the language


siddhantk232:
```FASTN.ftd
-- import: fastn

-- fastn.package: siddhant.xyz
default-language: en
translation-en: siddhant.xyz/i18n/en
translation-hi: siddhant.xyz/i18n/hi
```

```
-- ftd.text: $lang.title
```

The `en.ftd` and `hi.ftd` file have the `title` variable.

Changing the value of `default-language` to "hi" does work and the page renders
text in hindi. But, the cookie value `fastn-lang` has no effect.

PR #2163, opened 2025-06-30
fix for #2161

siddhantk232:
- **minor: clean trace output a little**
- **fix(i18n): Resolve requested language before auto processing imports**
- **fix: Make `set_current_language` callable from ftd**


siddhantk232:
I'll write docs for this after the test passes.

amitu (approved):

PR #2165, opened 2025-07-01
WIP fastn <package_name> command to launch a native window

siddhantk232:
- **feat: Add fastn <package_name> command**
- **feat: Tauri setup and some refactor**


fastn/src/lib.rs
```diff
@@ -0,0 +1,505 @@
...
+    // Launch UI with the default package
+    // TODO: use lets-os for the default pkg
+    let slug = "design-sytem";
+    fastn_ui(slug).await.expect("Failed to launch fastn UI");
+}
+
+#[tracing::instrument]
+pub async fn fastn_ui(slug: &str) -> fastn_core::Result<()> {
+    let pkg_dir = {
+        let mut data_dir = dirs::data_dir().unwrap();
```
amitu, line 26: Need some note on this. Where are these packages getting cloned? Can I run more than once instance of a package, eg checkout them twice, as each checkout will have its sqlite file for example. And if our file upload feature uploads files in the same folder, the two instances will have different files also going forward. Our editor may also modify these files etc. 

Actually for every package we should have a template package that should be checked out, as that is what most people want to do.

fastn/src/tauri.rs
```diff
@@ -0,0 +1,33 @@
...
+                        .level(log::LevelFilter::Info)
+                        .build(),
+                )?;
+            }
+
+            let webview = app
+                .get_webview_window("main")
+                .expect("main label is created by tauri");
+
+            let url = format!("http://localhost:{fastn_serve_port}")
```
amitu, line 21: Over my years of experience I have come to prefer 127.0.0.1. Either that or localhost should be the same. But sometimes it is not the same (/etc/hosts entry, people are funny). And every time they are not the same I wish I should have used 127.0.0.1. We can not assume every single place fastn is installed has not done anything funny in their /etc/hosts, and we know for sure not matter whats in their /etc/hosts, 127.0.0.1 will always be what you want, so 127.0.0.1 is the preferred always over localhost.

amitu:
> `fastn [OPTIONS] <package_name>`

Should be:

> `fastn [OPTIONS] [<package_name>]`

(`<package_name>` is optional, and defaults to lets-os). How do we update our help text to make this clear?

fastn/web_assets/index.html
```diff
@@ -0,0 +1 @@
+<h1>Hello fastn!</h1>
```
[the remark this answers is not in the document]
siddhantk232: We don't. I'll remove it.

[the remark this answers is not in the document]
siddhantk232: done in https://github.com/fastn-stack/fastn/pull/2165/commits/6f85671f7f4c435db23dac161776cca6d1de423f

fastn/src/tauri.rs
siddhantk232: done in https://github.com/fastn-stack/fastn/pull/2165/commits/6f85671f7f4c435db23dac161776cca6d1de423f

siddhantk232:
I am focusing on running the current version as a GUI only app for mac and android. This part has a lot of unknowns (does fastn-core even compile for android/arm?) so it's better to do this first.

amitu:
I guess it's the same issue or related to fastn not working in dev builds with stack overflow error. We really have to solve this someday. Not sure how to debug tho.

siddhantk232:
> I guess it's the same issue or related to fastn not working in dev builds with stack overflow error. We really have to solve this someday. Not sure how to debug tho.

Yes! After writing this comment I noticed that the tauri cli is building the crate in debug mode when I did `cargo tauri android dev`. There's a `--release` option, that'll probably work.

amitu:
Can we find out if `fastn` works on Android?

siddhantk232:
> Can we find out if `fastn` works on Android?

https://youtu.be/BT9QuvR48vc It's working :)

